PR #81, opened 2015-01-14
Update stable to master to catch prior to unified agent changes

ferventcoder:
```
* master: (23 commits)
  (PUP-1802) Use xslt filter to skip ruby.exe in sys.wxs
  (PUP-1802) Puppet agent should use ruby.exe directly
  (PUP-3471) Do not expand path variables
  (PUP-3003) Exclude util and patch files
  (PUP-3003) Exclude unnecessary files from MSI
  (maint) bumping 3.7.x yaml facter version to 2.2.0
  (PUP-3315) Ask for master's FQDN
  (PUP-2984) Remember x86 / x64 custom install dirs
  (PUP-2984) Remove x64 Program Files Path
  (PUP-2986) Add MIT license for Unknownr
  (PUP-3006) Block x64 installer on Windows 2003
  (PUP-390) Append -x64 to msi name
  (PUP-390) arch is more correct terminology over platform
  (maint) Store install location in uninstall key
  (PUP-390) Ensure x86/x64 remembered properties
  (PUP-390) Provide Puppet 3.7 config files
  (PUP-390) Add Windows x64 support to MSI
  (PE-4699) Tighten regex slightly
  Whitespace cleanup
  (PE-4699) Ensure PEVERSION constant is defined
  ...
```


joshcooper:
Does this need to wait until after puppet 3.7.4? I don't remember how that works

ferventcoder:
@joshcooper As long as we don't accept any PRs for master we should be fine. But I think everything is pinned to a specified commit.

ferventcoder:
This is probably generally safe, 3.7.3 is using master or a pin. I don't even know that we use stable anywhere yet. @haus @melissaanne could confirm this.

haus:
From puppet build_defaults:
```
build_msi:
  puppet_for_the_win:
    ref: '4eb71b5b063f611eb447d561d51481831a66b5dd'
    repo: 'git://github.com/puppetlabs/puppet_for_the_win.git'
```

joshcooper:
What about a security/patch release of PE?

ferventcoder:
Pretty sure PE is either pinned to master or a sha. I'm saying I don't think we currently use stable. I set this up awhile back for the idea we would move to  stable/master approach and I don't think we have yet.

haus:
@joshcooper that's still pinned in jenkins to a sha

melissaanne:
@joshcooper 
```
if [ "$PE_VERSION" = "3.3" ] ; then
  WIN_SHA=c0da30c929297ca4c83ce8f79e5c9b598f3562c3
elif [ "$PE_VERSION" = "3.7" ] ; then
  WIN_SHA=4eb71b5b063f611eb447d561d51481831a66b5dd
else
  WIN_SHA=origin/master
fi
```

Iristyle:
I know that we track the SHA refs in the other repos, whether it's [`build_defaults.yaml`](https://github.com/puppetlabs/puppet/blob/master/ext/build_defaults.yaml#L17-L31) for FOSS or via Jenkins job config for PE + friends (which will be redone for PE.next)...

I think it would be a bit nicer to have a human readable moniker that serves as a marker to help line things up so we can do something like:

```yaml
build_msi:
  puppet_for_the_win:
    ref: 'unified_4.0.0'
    repo: 'git://github.com/puppetlabs/puppet_for_the_win.git'
```

AFAIK, we're pretty good about not rewriting tags, so they should be safe enough for this process IMHO, and then we have a reference on both sides of the fence.

Given the Jenkins snippet above, c0da30c929297ca4c83ce8f79e5c9b598f3562c3 is tagged `PE_3.3` and 4eb71b5b063f611eb447d561d51481831a66b5dd is tagged `PE_3.7` or similar.  

It looks like we used to do this at one point @haus - any reason we stopped?

Is stable / master for this completely new installer work sufficient?  I generally don't advocate more branches since `mo' branches, mo' problems`, and I think we can branch off of an old SHA if we need to support old installers, etc... but just want to make sure we don't miss anything / make it difficult on ourselves down the road.

haus:
@Iristyle no reason

Iristyle:
FYI @ferventcoder / @haus / @melissa / @joshcooper - I pushed some tags up based on what I found in puppetlabs/puppet and what was mentioned above.  It would be great if we could add more tags for older versions if we have them handy, but it seems that might be difficult given this stuff is strewn through historical Jenkins jobs:

```
[puppet_for_the_win] git tag 3.7.3 4eb71b5b063f611eb447d561d51481831a66b5dd
[puppet_for_the_win] git tag 3.7.2 f4f74b1c8d8792089d1d3328b01f5ff075771eb0
[puppet_for_the_win] git tag 3.7.1 c61dda253b09bd855488f7b6dd7332fdf84d3039
[puppet_for_the_win] git tag 3.7.0 c61dda253b09bd855488f7b6dd7332fdf84d3039
[puppet_for_the_win] git tag PE_370 4eb71b5b063f611eb447d561d51481831a66b5dd
[puppet_for_the_win] git tag PE_330 c0da30c929297ca4c83ce8f79e5c9b598f3562c3
```

PR #83, opened 2015-01-14
(PUP-3842) Puppet should install MCO on FOSS

ferventcoder:
Add MCollective to be installed as part of the puppet agent
installer MSI. Previously MCO would only be installed if we were
building PuppetEnterprise.MSI. This removes the limitation and
additionally does the following:

- renames service from `pe-mcollective` to `mcollective`
- sets the intial account/password structure to the same as Puppet
which still defaults to `LocalSystem` and Empty password.


ferventcoder:
One more fixup needs to be done with this that I found (the environment.bat.erb file).

haus:
:+1:

PR #84, opened 2015-01-14
(PUP-3843) Remove branding

ferventcoder:
This removes PE specific branding. It does make the observation that
version is now preferred to the same way that PE formerly did it where
you pass in `PE_VERSION_STRING`. However that has been renamed to
`AGENT_VERSION_STRING` to more accurately reflect what it is used for.

This is still waiting on what happens with `puppet.conf` `graph`
setting in PE.

This depends on the work that went into https://github.com/puppetlabs/puppet_for_the_win/pull/83 and should be evaluated AFTER it.

PR #85, opened 2015-01-15
(PUP-3854) Add cfacter to build automation

melissa:
This is still in active development, I wanted an easy way to communicate the changes I'm currently making for this.


melissaanne:
Test job at http://jenkins-release.delivery.puppetlabs.net/job/MelissaBuildsMSI/

melissaanne:
This likely will have to be rebased once the work Rob is doing gets merged in.

melissaanne:
Successful build at http://jenkins-release.delivery.puppetlabs.net/job/MelissaBuildsMSI/8/

conf/windows/stage/bin/cfacter_interactive.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+echo Running CFacter on demand ...
+cd "%~dp0"
+call cfacter.exe %*
```
MikaelSmith: I'm not sure how this works, as I don't think we source environment.bat first.

conf/windows/stage/bin/cfacter.bat
```diff
@@ -0,0 +1,6 @@
+@echo off
+SETLOCAL
+
+call "%~dp0environment.bat" %0 %*
+
+ruby -S -- "%SCRIPT_NAME%" %*
```
MikaelSmith: This file probably isn't needed, because it's looking for cfacter.rb to run.

conf/windows/stage/bin/cfacter_interactive.bat
melissaanne: We definitely don't. I just realized that... I'm changing things up a little to make sure that does happen

melissaanne: It did definitely work though, by running `run_cfacter_interactive.bat`

wix/puppet.wxs
```diff
@@ -265,6 +273,16 @@
               </Component>
             </Directory>
           </Directory>
+          <Directory Id="CFacterAppData" Name="cfacter">
+            <Directory Id="CFactsDotD" Name="cfacts.d">
+              <Component Id="PuppetInstallerCFacts" Permanent="yes" Guid="4E7C234B-2837-4459-A20E-9503A922FCB9">
```
melissaanne: I just copied the Guid from facter. I have no idea what this should be. Same with the above line

conf/windows/stage/bin/cfacter.bat
```diff
@@ -0,0 +1,6 @@
+@echo off
+SETLOCAL
+
+call "%~dp0environment.bat" %0 %*
+
+cfacter.exe %*
```
melissaanne: In facter.bat, this line is `ruby -S -- "%SCRIPT_NAME%" %*`. We're using our ruby to call facter, but what does `%SCRIPT_NAME%` evaluate to? Is that just calling `C:\Program Files\Puppet Labs\Puppet\facter\bin\facter`?

wix/puppet.wxs
ferventcoder: You just make up a GUID - I tend to use the online guid generators.

ferventcoder: You don't want any to be the same.

conf/windows/stage/bin/cfacter.bat
MikaelSmith: You could find out by adding `echo "%SCRIPT_NAME%"`.

wix/puppet.wxs
joshcooper: See http://wixtoolset.org/documentation/manual/v3/howtos/general/generate_guids.html

conf/windows/stage/bin/cfacter.bat
joshcooper: `%SCRIPT_NAME%` in this case will be `cfacter`. The logic comes from https://github.com/melissaanne/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat.erb#L12-L15, which extracts the filename of the batch file it's being called from, and strips the extension.

I prefer that the bat file be explicit about what it's executing, so :+1: to calling `cfacter.exe` directly.

One minor thing, we probably want to specify the path to `cfacter.exe`, either `.\cfacter.exe` or `%~dp0cfacter.exe`. The `%~dp0` macro returns the directory relative to the batch file being executed. That way things work no matter what your current working directory is. Also the macro always contains a trailing backslash, so even though it looks like it's missing a backslash, it actually does the right thing.

joshcooper: Actually do we need this batch file? puppet and facter have one, because we have to set the ruby `$LOAD_PATH`, and execute `ruby.exe`.

We could have `cfacter_interactive.bat` just execute `cfacter.exe` directly. But perhaps we do need this batchfile in order for cfacter to resolve ruby-based facts, e.g. `rubyplatform`, pluginsync'ed and external facts?

conf/windows/stage/bin/cfacter_interactive.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+echo Running CFacter on demand ...
+cd "%~dp0"
+call cfacter.bat %*
```
joshcooper: this should probably be `.\cfacter.bat` or `.\cfacter.exe` depending on the result of https://github.com/puppetlabs/puppet_for_the_win/pull/85/files#r23056708

melissaanne: Is there a reason this isn't `call "%~dp0cfacter.bat" %*`?

melissaanne:
@joshcooper @MikaelSmith I've updated this PR. It will likely need to be rebased again once https://github.com/puppetlabs/puppet_for_the_win/pull/84 is merged, but I've addressed all your comments.

ferventcoder:
@joshcooper @melissaanne `environment.bat.erb` could technically become just `environment.bat` in my PR since there is technically no conditional logic in it anymore but I didn't make that adjustment yet. I thought I would follow it on later with a fixup.

wix/puppet.wxs
```diff
@@ -263,6 +271,16 @@
               </Component>
             </Directory>
           </Directory>
+          <Directory Id="CFacterAppData" Name="cfacter">
+            <Directory Id="CFactsDotD" Name="cfacts.d">
+              <Component Id="PuppetInstallerCFacts" Permanent="yes" Guid="1FD82022-CC2B-407B-86D5-7F14B5927C68">
```
melissaanne: @ferventcoder, @Iristyle, @joshcooper  So I'm getting `C:\Users\jenkins\workspace\puppet-msi-2015-01-20-16-13-51-16fd6ad\ARCH\x86\label\win-build\wix\puppet.wxs(276) : error LGHT0204 : ICE21: Component: 'PuppetInstallerCFacts' does not belong to any Feature.` when I run this. Like I mentioned earlier, I basically just copied facter for the contents here (however the Guid is unique). I'm going to investigate this further, but in the meantime, do you have any suggestions?

melissaanne:
This is dependent on https://github.com/puppetlabs/packaging/pull/371

conf/windows/stage/bin/cfacter.bat
```diff
@@ -0,0 +1,6 @@
+@echo off
+SETLOCAL
+
+call "%~dp0environment.bat" %0 %*
+
+"%~dp0cfacter.exe" %*
```
melissaanne: It actually looks like it's looking for cfacter.exe in `C:\Program Files\Puppet Labs\Puppet\bin\cfacter.exe`. Since calling the environment sets `PATH`, our cfacter.exe should be executable. Either that, or we should call `%CFACTER_DIR%\bin\cfacter` to ensure we're calling the correct thing (if that actually works, which I will check)

conf/windows/stage/bin/cfacter_interactive.bat
```diff
@@ -0,0 +1,5 @@
+@echo off
+echo Running CFacter on demand ...
+cd "%~dp0"
+call .\cfacter.bat %*
```
melissaanne: So, I asked this earlier, but I think the question got lost in the updates I made. But, is there a reason this is on two lines, and not `call "%~dp0cfacter.bat" %*`?

melissaanne:
I'm still getting the above compilation error, but I do have successful builds at http://builds.puppetlabs.lan/puppet/16fd6ad7ea19b6922fc55c611eb7d27e371878ee/artifacts/windows/ the jenkins job that built them is http://jenkins-release.delivery.puppetlabs.net/job/puppet-msi-2015-01-20-18-00-06-16fd6ad/1/

melissaanne:
@ferventcoder also, as to your comment above, I added a commit that removed the erb template. Check out https://github.com/melissaanne/puppet_for_the_win/commit/a977a668a59ace39e1a45e49b4f5f37d41a9eff0 for what I changed

wix/puppet.wxs
melissaanne: I was missing the line added at `:793` :)

melissaanne: Yup, I'm no longer seeing that error http://jenkins-release.delivery.puppetlabs.net/job/puppet-msi-2015-01-20-18-39-59-e773548/ARCH=x64,label=win-build/1/console

tasks/windows/windows.rake
```diff
@@ -138,7 +128,14 @@ namespace :windows do
     APPS.each do |name, config|
       if not File.exists?("downloads/#{name}")
         Dir.chdir "#{TOPDIR}/downloads" do
-          sh "git clone #{config[:repo]} #{name}"
+          if config[:path]
+            sh "curl -O #{config[:path]}/#{config[:archive]}"
+            if config[:archive] =~ /^.*\.zip$/
+              sh "unzip #{config[:archive]} -d #{name}"
+            end
+          else
```
melissa: This assumes the new format will be https://github.com/puppetlabs/puppet/pull/3482/files

conf/windows/stage/bin/run_cfacter_interactive.bat
```diff
@@ -0,0 +1,2 @@
+@echo Running cfacter on demand ...
+@"%~dp0..\sys\tools\bin\elevate.exe" "%~dp0cfacter_interactive.bat"
```
haus: It seems like this will display running cfacter on demand twice, which is odd.

tasks/windows/windows.rake
```diff
@@ -147,6 +144,7 @@ namespace :windows do
   task :checkout => :clone do
     APPS.each do |name, config|
       Dir.chdir "#{TOPDIR}/downloads/#{name}" do
+        next unless config[:repo]
```
haus: This can be moved up a line, outside the chdir block.

melissa: Very true

conf/windows/stage/bin/cfacter_interactive.bat
Iristyle: The only thing that I can think of is that the current working directory is important for some reason (i.e. if things are loaded relative to CWD).  I'm not qualified to answer that, but perhaps @MikaelSmith or @peterhuene could say for sure.

MikaelSmith: It's only important if the user is specifying arguments and trying to use relative paths for custom-dir or external-dir.

wix/puppet.wxs
```diff
@@ -89,6 +89,14 @@
               WorkingDirectory="bin">
               <Icon Id="FacterShortcutIcon" SourceFile="conf\windows\stage\misc\puppetlabs.ico"/>
             </Shortcut>
+            <Shortcut Id="CFacterShortcut"
```
melissa: If we don't include this cfacter stuff, will it not install the cfacter.bat, etc. files? I'm thinking it might be fun to make this a template based on what gets passed in with the yaml file. But that could also be more painful that it's worth.

wix/puppet.wxs
```diff
@@ -261,6 +269,16 @@
               </Component>
             </Directory>
           </Directory>
+          <Directory Id="CFacterAppData" Name="cfacter">
+            <Directory Id="CFactsDotD" Name="cfacts.d">
```
melissa: Should this be facts.d, rather than cfacts?

MikaelSmith: Yes.

ferventcoder: Which means we don't need this then ?)

MikaelSmith: Right, we don't use a separate custom facts directory for cfacter.

conf/windows/stage/bin/run_cfacter_interactive.bat
ferventcoder: One proper case and one lower case.

melissa:
Updated. The latest builds are available at http://builds.puppetlabs.lan/puppet/c7c281ae875ee8448e5962be280fadde6445bb17/artifacts/windows/

melissa:
packages build with this automation, containing cfacter 0.3.0 http://builds.puppetlabs.lan/puppet/94f021bdb8686273959d92a2f2689974d04afd24/artifacts/windows/

ferventcoder:
Yay! So this is all ready to go now?

conf/windows/stage/bin/run_cfacter_interactive.bat
ferventcoder: @melissa can you fix this one up?

melissa: Ah, I didn't realize it was something you wanted changed :)

melissa:
@ferventcoder this should be good

conf/windows/stage/bin/run_cfacter_interactive.bat
ferventcoder: No worries, having the same message twice is a bit of an annoyance. Something we could avoid. :)

ferventcoder:
\o/

PR #86, opened 2015-01-16
(PUP-2638) create the $vardir/state/graphs folder

ferventcoder:
Running `puppet apply` before `puppet agent` with `graph` set to true
will cause the catalog application to fail because the required
directory is missing. Previously you would need to create
the directory manually or call `puppet agent -t` at least once, even if
to a non-existent master.

This also picks up at least one important maint commit - 

```
(maint) Remove Puppet Master compatibility message

This removes the notice of what version of puppet masters the product
will be compatible because it could be a Puppet Master, Puppet Server,
and those various versions. Honestly that information can be looked up
elsewhere, it doesn't provide much value in the installer.
```

### DIFF
--- a/wix/localization/puppet_en-us.wxl
+++ b/wix/localization/puppet_en-us.wxl
@@ -256,8 +256,8 @@
 
     <String Id="WelcomeDlg_Title" Overridable="yes"><!-- _locID_text="WelcomeDlg_Title" _locComment="WelcomeDlg_Title" -->[ProductName] Setup</String>
     <String Id="WelcomeDlgBitmap" Overridable="yes"><!-- _locID_text="WelcomeDlgBitmap" _locComment="WelcomeDlgBitmap" -->WixUI_Bmp_Dialog</String>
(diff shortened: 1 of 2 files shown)


joshcooper:
Updating the MSI will work for 99% of the cases, but it seems like the more correct thing would be to modify puppet so that it creates the necessary directories prior to using them using the  `Puppet.settings.use(:graphing)` construct. For example, changing the MSI won't help if the user runs `puppet apply ... --vardir <somewhere>`

ferventcoder:
True. If we decide not to move forward with this, we'll want to get the other commits in still.

Iristyle:
Agreed with Josh - the setup should be part of the runtime so it affects all platforms, not the installer.

ferventcoder:
I'm not disagreeing at all. Moving it to that means it falls back with other priorities and out of our sprint though.

ferventcoder:
Alright fair enough. I will cherry pick the other two commits into master and close this if everyone agrees?

Iristyle:
Commits 0d97288 and e67ef14 have been cherry-picked to master.

ferventcoder:
Sweet, was holding this open until I or someone else addressed that. :)

PR #87, opened 2015-01-20
(MCO-548) Discontinue use of MCO daemon.bat

ferventcoder:
Currently when mcollective runs as a service on Windows, it executes a
batch script in a cmd.exe process. This adds a layer of unnecessary
complexity when we could call our vendored rubyw directly with the
necessary arguments for the service to run.

This switches over to using rubyw.exe directly and updates filters.xslt
to exclude rubyw.


wix/puppet.wxs
```diff
@@ -212,6 +212,27 @@
...
+                      Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                      Description="Puppet Labs server orchestration framework"
+                      DisplayName="Marionette Collective Server"
+                      Interactive="no"
+                      Name="mcollective"
+                      Start="demand"
+                      Type="ownProcess"
+                      ErrorControl="normal"
+                      Vital="yes" 
+                      Arguments="-I&quot;[INSTALLDIR]puppet\lib;[INSTALLDIR]facter\lib;[INSTALLDIR]hiera\lib;[INSTALLDIR]mcollective\lib;[INSTALLDIR]mcollective_plugins&quot; -rubygems &quot;[INSTALLDIR]mcollective\bin\mcollectived&quot; --daemonize"
```
ferventcoder: Sooo long.... ugh.

ferventcoder: @joshcooper thoughts?

ferventcoder: @Iristyle thoughts as well ?

Iristyle:
I assume we're also going to want a PR to remove  https://github.com/puppetlabs/marionette-collective/blob/master/ext/windows/daemon.bat

wix/puppet.wxs
Iristyle: For reference, the original `daemon.bat` this replaces looks like:

```batch
@echo off
SETLOCAL

call "%~dp0..\bin\environment.bat" %0 %*

rubyw -rubygems "%MCOLLECTIVE_DIR%\bin\mcollectived" %*
```

And the original `environment.bat` looks like:

```batch
SET BASEDIR=%~dp0..
SET BASEDIR=%BASEDIR:\bin\..=%

SET SERVER_CONFIG=%BASEDIR%\etc\server.cfg
SET CLIENT_CONFIG=%BASEDIR%\etc\client.cfg

SET MCOLLECTIVED=%BASEDIR%\bin\mcollectived
SET MC_STARTTYPE=manual
REM SET MC_STARTTYPE=auto

SET PATH=%BASEDIR%\bin;%PATH%

SET RUBYLIB=%BASEDIR%\lib;%RUBYLIB%
SET RUBYLIB=%RUBYLIB:\=/%

SET RUBY="ruby"
```

Iristyle: However, note that the `environment.bat` from above is not the one used by MCO after the installer has been built.  Instead, the [`environment.bat`](/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat.erb) called is the one that's part of this repository (and generated from `environment.bat.erb`).

Iristyle: So one thing that I'm concerned about is that previously we manipulated `PATH` to ensure that all of the tools we might need to launch from MCO are available.  A common use case for MCO is to trigger an on-demand Puppet run to apply a new manifest that was just modified on the server.

In the batch file, we're setting `PATH` via:

```batch
SET PATH=%PUPPET_DIR%\bin;%FACTER_DIR%\bin;%HIERA_DIR%\bin;%MCOLLECTIVE_DIR%\bin;%PL_BASEDIR%\bin;%PL_BASEDIR%\sys\ruby\bin;%PL_BASEDIR%\sys\tools\bin;%PATH%
```

In this modified service installation, since we're going straight to [`mcollectived`](https://github.com/puppetlabs/marionette-collective/blob/master/bin/mcollectived) and not through a `daemon.rb` like we do on the Puppet side (that ends up triggering `environment.bat`), the `PATH` is no longer configured in the same manner.

ferventcoder:
We'll eventually want to delete the puppet daemon.bat as well. :)

PR #88, opened 2015-01-28
(PUP-3845) Track versions of components

ferventcoder:
Track versions of components in the MSI and install so that folks can
see exactly what went into the package. Create the file at
stagedir/misc just before heating the components so it gets baked into
the MSI and subsequently installed in
`INSTALLDIR/Puppet/misc/versions.txt`.

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -220,7 +220,17 @@ namespace :windows do
     FileUtils.rm_rf(["stagedir/mcollective/lib/mcollective/vendor/json", "stagedir/mcollective/lib/mcollective/vendor/load_json.rb"])
   end
 
-  task :wxs => [ :stage, 'wix/fragments' ] do
+  task :track_versions do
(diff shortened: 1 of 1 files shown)


ferventcoder:
@haus @melissa this should be relatively simple.

haus:
@ferventcoder looks sane to me. i was wondering about += vs <<, but I'm +1 either way.

ferventcoder:
oh good point. I'd imagine either work work. Just was using an old string concat I knew off the top of my head. Can switch it over to `<<` if there are concerns with using it.

haus:
I have no concerns, just the \n and += look awkward to me. pretty sure with << you don't need to add the \n.

haus:
like i said, :+1: either way.

ferventcoder:
I can switch it over. :)

ferventcoder:
We'll leave it this way. Looks like `<<` would still require the line feed

haus:
cool.

PR #89, opened 2015-02-04
(maint)(packaging) Updates for MCO 2.8 release

shrug:
the marionette-collective 2.8 release moved the
core plugins into the lib/ directory instead
of a separate plugins directory.

This removes the dependency on the plugins/
directory


haus:
:+1: Seems sane

waynr:
This is my first time seeing this repo and I am not windoze or aio aware so I can't really review/approve. I can, however, approve of the repo name!

melissa:
:+1: It looks like you got everything that was needed. @ferventcoder have you had a chance to look over this?

ferventcoder:
Looks sane. There may be another spot to clean up but :+1:

ferventcoder:
@Iristyle look good to you?

Iristyle:
So as I understand it, this was partially due to changes necessary to run MCO as a gem, correct @ploubser / @richardc?

Are we at all concerned about the migration path for existing users?  Or is our recommendation going to be **reinstall all your plugins**?  It would be possible for the installer to move things from `mcollective_plugins` to the new location, but that's a slippery slope given downgrades, etc.

I know very little about MCO, but I know that in the past to test things out, I've used at least 2 different MCO plugins.  I have no idea how common it is for our end users to use plugins?

- https://github.com/puppetlabs/mcollective-facter-facts
- https://github.com/thesilentpenguin/shellcmd-agent


My gut says we might want to at least bridge the old and new worlds for some period of time... but maybe you guys have already had that discussion?

/cc @kylog @joshcooper

haus:
@Iristyle are you referring to the path changes or to the plugins themselves moving? The packaging should take care of moving any plugins that are part of the package, and leave any old plugins behind. 
We probably do want to make sure that the mco libdir setting includes the old plugindir to ensure that if folks have installed any plugins that they continue to work. One way would be to leave the RUBYLIB overrides in place. Another way would be via the mco server config libdir setting.

joshcooper:
Are user-plugins (not provided by the package) installed in `C:\Program Files\...\mcollective_plugins` or in `C:\ProgramData\...`?

haus:
@joshcooper I don't know. We don't currently package them for windows. Looking at the mco pluginpackager code i only see support baked in for rpm and deb, so I don't know how users would install them. For PE, here is where extra plugins are installed: https://github.com/puppetlabs/puppetlabs-puppet_enterprise/blob/3.7.x/manifests/params.pp#L102-L104

richardc:
@IristyleI think @haus already answered your question - nothing breaks for existing users with custom plugins, they still have to be put somewhere, and $libdir still needs to be pointed at them.

Custom plugins shouldn't be installed to the old path, but even if they were then the packaging system should be smart enough to know that the files don't belong to it, so should not be removed on upgrade.

haus:
PE on windows should be fine here. it is already setting libdir in the template to both core dir and plugindir.

conf/windows/stage/bin/environment.bat
```diff
@@ -21,12 +21,11 @@ SET FACTER_DIR=%PL_BASEDIR%\facter
 SET CFACTER_DIR=%PL_BASEDIR%\cfacter
 SET HIERA_DIR=%PL_BASEDIR%\hiera
 SET MCOLLECTIVE_DIR=%PL_BASEDIR%\mcollective
-SET MCOLLECTIVE_PLUGIN_DIR=%PL_BASEDIR%\mcollective_plugins
 
 SET PATH=%PUPPET_DIR%\bin;%FACTER_DIR%\bin;%CFACTER_DIR%\bin;%HIERA_DIR%\bin;%MCOLLECTIVE_DIR%\bin;%PL_BASEDIR%\bin;%PL_BASEDIR%\sys\ruby\bin;%PL_BASEDIR%\sys\tools\bin;%PATH%
 
 REM Set the RUBY LOAD_PATH using the RUBYLIB environment variable
-SET RUBYLIB=%PUPPET_DIR%\lib;%FACTER_DIR%\lib;%HIERA_DIR%\lib;%MCOLLECTIVE_DIR%\lib;%MCOLLECTIVE_PLUGIN_DIR%;%RUBYLIB%;
+SET RUBYLIB=%PUPPET_DIR%\lib;%FACTER_DIR%\lib;%HIERA_DIR%\lib;%MCOLLECTIVE_DIR%\lib;%RUBYLIB%;
```
ferventcoder: This may need to stay based on the end of this conversation

ferventcoder: Or not, I don't completely follow the conversation above.

ferventcoder: @haus from your comment https://github.com/puppetlabs/puppet_for_the_win/pull/89#issuecomment-73159532 - it sounds like this needs to stay...

haus: @ferventcoder PE sets libdir in server.cfg for MCO to include this path, so removing it here *shouldn't* break anything.

Iristyle:
After some more discussion, the changes here are probably OK given existing MCO installs that were configured via Puppet module will have their `site.cfg` pointed at the appropriate directories on Windows already.  Therefore changes to the batch file shouldn't have a material impact on where plugins are found based on previous configuration:

```puppet
$mco_plugin_libdir  = "${mco_etc}/plugins;${::env_windows_installdir}/mcollective_plugins"
```

I assume that the next iteration of the module will do something like this:

```puppet
$mco_plugin_libdir  = "${mco_etc}/plugins;${::env_windows_installdir}/lib;${::env_windows_installdir}/mcollective_plugins"
```

But that's out of the scope of this PR.

PR #90, opened 2015-02-05
(maint) puppet agent updates

melissa:
This pull request adds in updates that are needed to build the puppet-agent msi on a vmpooler image. We are no longer using the puppetwinbuilder tarball to provide the necessary packages. Those are instead installed via chocolatey. Because of that, there are a few differences in how we invoke commands.

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -134,7 +134,7 @@ namespace :windows do
               begin
                 sh "unzip #{config[:archive]} -d #{name}"
               rescue
-                sh "C:\\ProgramData\\chocolatey\\bin\\7za.exe x -r -tzip -o#{name} #{config[:archive]}"
+                sh "7za x -r -tzip -o#{name} #{config[:archive]}"
               end
               sh "rm #{config[:archive]}"
             end
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -113,7 +113,7 @@ namespace :windows do
   CONFIG = YAML.load_file(ENV["config"] || "config.yaml")
   APPS = CONFIG[:repos]
   ENV['ARCH'] = ENV['ARCH'] || 'x86'
-  ENV['PKG_FILE_NAME'] = ENV['PKG_FILE_NAME'] || "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi"
+  ENV['PKG_FILE_NAME'] = (ENV['PKG_FILE_NAME'] || "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi").gsub(/\s+/, "")
 
   task :clean_downloads => 'downloads' do
     FileList["downloads/*"].each do |repo|
@@ -300,7 +300,7 @@ namespace :windows do
     if not ENV['AGENT_VERSION_STRING']
       puts "Warning: AGENT_VERSION_STRING is not set in the environment.  Defaulting to 1.0.0"
       ENV['AGENT_VERSION_STRING'] = '1.0.0'
-      ENV['PKG_FILE_NAME'] = "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi"
+      ENV['PKG_FILE_NAME'] = "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi".gsub(/\s+/, "")
     end
     Rake::Task["windows:msi"].invoke
   end
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -131,7 +131,11 @@ namespace :windows do
           if config[:path]
             sh "curl -O #{config[:path]}/#{config[:archive]}"
             if config[:archive] =~ /^.*\.zip$/
-              sh "unzip #{config[:archive]} -d #{name}"
+              begin
+                sh "unzip #{config[:archive]} -d #{name}"
+              rescue
+                sh "C:\\ProgramData\\chocolatey\\bin\\7za.exe x -r -tzip -o#{name} #{config[:archive]}"
+              end
               sh "rm #{config[:archive]}"
             end
           else


tasks/windows/windows.rake
```diff
@@ -131,7 +131,11 @@ namespace :windows do
           if config[:path]
             sh "curl -O #{config[:path]}/#{config[:archive]}"
             if config[:archive] =~ /^.*\.zip$/
-              sh "unzip #{config[:archive]} -d #{name}"
+              begin
+                sh "unzip #{config[:archive]} -d #{name}"
+              rescue
+                sh "C:\\ProgramData\\chocolatey\\bin\\7za.exe x -r -tzip -o#{name} #{config[:archive]}"
```
ferventcoder: This seems sane.

ferventcoder: It bothered me at first but I saw it was a backup.

melissa: Yeah, it's gross. Mainly, when I was trying to build this on a vmpooler without the puppetwinbuilder tarball, we don't have unzip on the box. Actually, I should be able to update the pathing so we don't have to be explicit about where this executable is.

Iristyle: Isn't the `puppetwinbuilder` zip required to build things?

Like you said, it includes `sys\git\bin\unzip.exe`.  It also includes `sys\misc\bin\7za.exe`.

I don't totally understand the circumstances you're describing above @melissa

tasks/windows/windows.rake
```diff
@@ -113,7 +113,7 @@ namespace :windows do
   CONFIG = YAML.load_file(ENV["config"] || "config.yaml")
   APPS = CONFIG[:repos]
   ENV['ARCH'] = ENV['ARCH'] || 'x86'
-  ENV['PKG_FILE_NAME'] = ENV['PKG_FILE_NAME'] || "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi"
+  ENV['PKG_FILE_NAME'] = (ENV['PKG_FILE_NAME'] || "puppet-agent-#{ENV['AGENT_VERSION_STRING']}-#{ENV['ARCH']}.msi").gsub(/\s+/, "")
```
Iristyle: What causes spaces in the name?

tasks/windows/windows.rake
melissa: https://github.com/puppetlabs/puppet-agent/pull/32 is what these changes are for. That adds in a ruby script that builds both cfacter and the puppet-agent msi. Just, because, apparently we need yet another way to build the msi. The puppetwinbuilder tar includes ruby, git, zip, rake, and some other packages needed for building. Instead of making these available via that tarball, we're installing them with chocolatey. Since we're calling this task through a series of environment changes (bash to cygwin to cmd.exe etc), preserving the environment has proved challenging.

tasks/windows/windows.rake
melissa: I'm not sure what's causing the additional spaces. I think it's something to do with how I'm propagating the environment variables through all the different levels of building

Iristyle: I would be in favor of tracking down the root cause of the issue, if for no other reason that to have a better understanding of what's going on (would be useful to add to commit message as well)

melissa: I'll keep playing with it and see what I can come up with

tasks/windows/windows.rake
```diff
@@ -131,7 +131,11 @@ namespace :windows do
           if config[:path]
             sh "curl -O #{config[:path]}/#{config[:archive]}"
             if config[:archive] =~ /^.*\.zip$/
-              sh "unzip #{config[:archive]} -d #{name}"
+              begin
+                sh "unzip #{config[:archive]} -d #{name}"
+              rescue
+                sh "7za x -r -tzip -o#{name} #{config[:archive]}"
```
ferventcoder: So this is good now? I'm good to merge if you are happy with this.

melissa: The 7za is good, but I have yet to dig into why we're seeing spaces in the environment variables we're passing through

ferventcoder: Not as worried about that...

ferventcoder: Should we create a followup ticket to address that and move this forward?

tasks/windows/windows.rake
ferventcoder: @Iristyle to answer your question - `puppetwinbuilder` is not required and RE is moving away from the use of it entirely.

tasks/windows/windows.rake
Iristyle: I think spaces making their way in seem indicative of an issue that should be addressed if possible.

ferventcoder: I agree, I just don't see it as a blocker for moving this PR.

tasks/windows/windows.rake
ferventcoder: I think there may be an issue in puppetlabs/packaging not stripping off the output from git describe somewhere.

ferventcoder: It's based on these comments https://github.com/puppetlabs/packaging/pull/375#issuecomment-73603543

haus: This is the output on osx. We call strip or chomp on the returned output from all of the git commands AFAICT.

```
irb(main):006:0> puts "'#{Pkg::Config.version}'"
'3.7.3-1450'
=> nil
irb(main):007:0> puts "'#{Pkg::Util::Version.git_describe}'"
'3.7.3-1450-g77c2213'
=> nil
irb(main):008:0> puts "'#{Pkg::Config.short_ref}'"
'77c2213'
=> nil
```

melissa: If anything, it's an issue with https://github.com/puppetlabs/puppet-agent/pull/32 These changes aren't needed when we build the msi with the packaging repo. My theory is the issue is with how they're getting passed in at https://github.com/melissa/puppet-agent/blob/ticket/master/build-cfacter-msi/bin/build-windows.rb#L107

ferventcoder:
Let's do this. We have a non-blocking issue here. @melissa do you want to file the ticket to followup?

melissa:
@ferventcoder https://tickets.puppetlabs.com/browse/RE-4004 I made it an RE ticket, since I wasn't sure what project it should belong to.

PR #91, opened 2015-02-12
(maint) Add the cfacter libdir to the ruby path

melissa:
We need to make sure this is available so that cfacter can load custom
facts, etc.


MikaelSmith:
:+1:

ferventcoder:
shiny

PR #92, opened 2015-02-12
(maint) use git describe --long when tag is HEAD

Iristyle:
- Git has some interesting behavior when using `git describe` when the
   current HEAD commit is a tag.  For instance, against the
   puppet-win32-ruby repository and git version 1.7.8.msysgit.0, the
   following behavior is observed, when performing a describe against
   93d40ef (which is tagged as 2.1.5.0-x86):

   > git describe
   > 2.1.5.0-x86
   > git describe --long
   > 2.1.5.0-x86-0-g93d40ef

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -56,9 +56,16 @@ def variable_define_flags
   flags.inject([]) { |a, (k,v)| a << "-d#{k}=\"#{v}\"" }.join " "
 end
 
-def describe(dir)
+def describe(dir, style = :long)
   @git_tags ||= Hash.new
-  @git_tags[dir] ||= Dir.chdir(dir) { %x{git describe}.chomp }
+  @git_tags[dir] ||= Dir.chdir(dir) do
+    {
+      :long => %x{git describe --long}.chomp,
+      :short => %x{git describe}.chomp
+    }
+  end
+
+  @git_tags[dir][style]
 end
 
 def cp_p(src, dest, options={})
@@ -258,7 +265,7 @@ namespace :windows do
     msg = 'Could not parse git-describe annotated tag for MCollective'
     match_data=[]
     @version_regexps.find(lambda { raise ArgumentError, msg }) do |re|
-      match_data = (describe 'downloads/mcollective').match re
+      match_data = (describe 'downloads/mcollective', :short).match re
     end
     mco_version="#{match_data[1]}.#{match_data[2]}.#{match_data[3]}." << (match_data[4] || 0).to_s
     modified = content.gsub("@DEVELOPMENT_VERSION@", "#{mco_version}")
--- a/wix/include/puppet.wxi
+++ b/wix/include/puppet.wxi
@@ -4,8 +4,12 @@
   <?define OurCompanyNameWord="PuppetLabs"?>
   <?define OurCommonProductNameWord="PuppetInstaller"?>
 
-  <!-- Component Versions.  We expect the build system to define these values
-       on the command line using `git describe` -->
+  <!--
+  <![CDATA[
+  Component Versions.  We expect the build system to define these values
+         on the command line using `git describe ––long`
+  ]]>
+  -->
   <?ifndef PuppetDescTag ?><?define PuppetDescTag="X.Y.Z-123-abcdef1" ?><?endif ?>
   <?ifndef FacterDescTag ?><?define FacterDescTag="X.Y.Z-456-fedcba9" ?><?endif ?>
 


Iristyle:
Having trouble giving this a thorough validation over VPN right now due to networking issues inside VM.

ferventcoder:
This looks appropriate for here.

ferventcoder:
@haus @melissa I'm good with this for informational sake in the versions.txt file, as long as it doesn't affect mco versioning in here.

ferventcoder:
@Iristyle I'll take a look and make sure that this does not affect mco versioning. We don't want Long when it is an actual release.

ferventcoder:
@Iristyle Yeah, that is not going to work. We'll need describe to take a parameter instead.

https://github.com/Iristyle/puppet_for_the_win/blob/maint/master/git-describe-long/tasks/windows/windows.rake#L261

haus:
@Iristyle why is just the tag bad? if versions.txt has just a tag, then you still know what was used to build it right?

Iristyle:
Not sure I understand the objection @ferventcoder.

versions.txt ends up looking like

```
cfacter PE_310-82-g4efecc4
facter 2.4.0-0-g8ff4fa5
hiera 1.3.4-0-g7d21886
mcollective 2.7.0-0-g07cb2f2
puppet 3.7.4-1347-gc8b9cfe
sys 2.1.5.0-x64-0-gb313f5b
```

haus:
@ferventcoder the version is used at build time to adjust mco's version, so i imagine if you type `mco --version` you'd get 2.7.0-0-g07cb2f2, which might be surprising to users.

haus:
er... @Iristyle

Iristyle:
I think it's a matter of precision @haus (even though we all know tags should never change).

If some of the included bits have SHAs, why not all of them?


I recently ran into a similar issue trying to figure what versions of things were included in the different RubyInstaller toolchain packages (that include gcc + mingw-w64).  It would have been enormously helpful to have SHAs present for all the included bits.  My 2c.

Iristyle:
@haus / @ferventcoder but when I looked at the regex definitions, they still looked fine, which is why I was asking why it's a problem.  Sorry, should have been more clear.

After building the MSI, and installing it.. you can see that the resultant patched `mcollective.rb` is fine:

```
C:\Program Files\Puppet Labs\Puppet\mcollective\lib>find "VERSION" mcollective.rb

---------- MCOLLECTIVE.RB
  VERSION="2.7.0.0"
    VERSION
```

haus:
@Iristyle the second `.0` shouldn't be there.

Iristyle:
@haus Ahhhhh... yes, I see the issue now (though not sure what that ends up impacting).

haus:
@Iristyle probably just docs reference lookups for users and mild confusion (they won't be able to find a version called 2.7.0.0 anywhere).

ferventcoder:
@Iristyle @haus I think you found what I was talking about. :)

Iristyle:
Actually @haus / @ferventcoder  ... I think the last `0` **is** supposed to be there per the line of code that does the capturing:


https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L263

```ruby
mco_version="#{match_data[1]}.#{match_data[2]}.#{match_data[3]}." << (match_data[4] || 0).to_s
```

I've added some code to ensure we grab the `short` (default) describe output, and the end result inside `mcollective.rb` is the same even upon confirming that the describe returns `2.7.0`.  `mcollective.rb` ends up with `2.7.0.0`

I added additional code to capture / cache both describes, but it's not strictly necessary given existing behavior.  I think that this should now be safe to merge.

Iristyle:
Note that the current YAML I'm using specifies:

```yaml
---
:repos:
  puppet:
    :ref: origin/master
    :repo: git://github.com/puppetlabs/puppet.git
  facter:
    :ref: 2.4.0
    :repo: git://github.com/puppetlabs/facter.git
  cfacter:
    :archive: 'cfacter-0.3.0-45-g71ebd28-x64.zip'
    :path: 'http://builds.puppetlabs.lan/cfacter/71ebd28c5570ee8c30a4c23f2ac4af1fa3ed7190/artifacts/windows/'
  hiera:
    :ref: 1.3.4
    :repo: git://github.com/puppetlabs/hiera.git
  sys:
    :ref: origin/2.1.x-x64
    :repo: git://github.com/puppetlabs/puppet-win32-ruby.git
  mcollective:
    :ref: 2.8.0
    :repo: git://github.com/puppetlabs/marionette-collective.git
```

And the `versions.txt` is written as:

```
cfacter PE_310-86-g175318d
facter 2.4.0-0-g8ff4fa5
hiera 1.3.4-0-g7d21886
mcollective 2.8.0-0-gd3c6dce
puppet 3.7.4-1351-g5895c0f
sys 2.1.5.0-x64-0-gb313f5b
```

Keep in mind of course that CFacter still doesn't have it's own `version` file, so it's picking up the `git describe --long` for the `puppet_for_the_win` repo.

haus:
@Iristyle the 4th dot was used in PE 3.x to indicate changes beyond the foss tag in the PE fork. That should no longer be necessary in the post PE 3.x world.

Iristyle:
@haus my point is that this PR didn't change the existing behavior around how the MCO version is written to `mcollective.rb`.

I'll open a separate PR for the issue you're describing.

haus:
@Iristyle fair and sounds good

Iristyle:
#94 opened to address 4th version field @haus

haus:
:+1:

PR #93, opened 2015-02-12
(maint) Update FOSS 4 YAML configs

Iristyle:
- Use Ruby 2.1.5 puppet-win32-ruby
 - Add Facter 2.4.0
 - Add MCO 2.8.0
 - Include CFacter refs


Iristyle:
FYI @ferventcoder / @melissa

PR #95, opened 2015-02-18
(PUP-4006) Non-Git dependencies use VERSION file

Iristyle:
- Previously, for extracted zips, such as CFacter, the versions.txt
   included a `git describe --long` for the directory.  This would
   actually traverse to a parent directory until it found one, in this
   case the puppet_for_the_win repository itself.  As a result, the
   versions.txt would contain the version of the puppet_for_the_win
   repo, instead of the actual dependency.

   For CFacter, a VERSION file was added as part of:
   https://github.com/puppetlabs/cfacter/commit/26654b3

   When the current download directory has a .git folder, indicating a
   Git repo, then use `git describe --long` as we always have.

   Otherwise, read the contents of the `VERSION` file

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -223,7 +223,9 @@ namespace :windows do
     version_tracking_file = 'stagedir/misc/versions.txt'
     content = ""
     FileList["downloads/*"].each do |repo|
-      content += "#{File.basename(repo)} #{describe repo}\n"
+      ver = Dir.exists?("#{repo}/.git") ?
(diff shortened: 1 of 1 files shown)


Iristyle:
Oops... PUP-4006 was still in doing @ferventcoder and not ready for merge yet.  

I guess I should have added a `DO NOT MERGE` to the subject.

ferventcoder:
No worries. It looked like it wasn't going to break anything the way it was done.

Iristyle:
#96 has the updated PR

PR #96, opened 2015-02-18
(PUP-4006) Non-Git dependencies use VERSION file

Iristyle:
- Previously, for extracted zips, such as CFacter, the versions.txt
   included a `git describe --long` for the directory.  This would
   actually traverse to a parent directory until it found one, in this
   case the puppet_for_the_win repository itself.  As a result, the
   versions.txt would contain the version of the puppet_for_the_win
   repo, instead of the actual dependency.

   For CFacter, a VERSION file was added as part of:
   https://github.com/puppetlabs/cfacter/commit/26654b3

   When the current download directory has a .git folder, indicating a
   Git repo, then use `git describe --long` as we always have.

   Otherwise, read the contents of the `VERSION` file

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -223,7 +223,22 @@ namespace :windows do
     version_tracking_file = 'stagedir/misc/versions.txt'
     content = ""
     FileList["downloads/*"].each do |repo|
-      content += "#{File.basename(repo)} #{describe repo}\n"
+      is_git = File.exists?("#{repo}/.git")
+      if is_git
+        version = describe(repo)
+      else
+        version_path = Dir["#{repo}/VERSION", "#{repo}/bin/VERSION"].first
+        if !version_path
+          abort "FAIL: #{repo} is not a Git repo, and is missing a VERSION file"
+        end
+
+        version = File.read(version_path).strip
+
+        if !version || version.empty?
+          abort "FAIL: #{repo} VERSION file missing version information"
+        end
+      end
+      content += "#{File.basename(repo)} #{version}\n"
     end
 
     File.open(version_tracking_file, "wb") { |f| f.write(content) }


tasks/windows/windows.rake
```diff
@@ -223,8 +223,13 @@ namespace :windows do
     version_tracking_file = 'stagedir/misc/versions.txt'
     content = ""
     FileList["downloads/*"].each do |repo|
-      ver = Dir.exists?("#{repo}/.git") ?
-        describe(repo) : File.read("#{repo}/VERSION").strip
+      is_git = Dir.exists?("#{repo}/.git")
+      if is_git
+        ver = describe(repo)
+      else
+        ver_path = ["#{repo}/VERSION", "#{repo}/bin/VERSION"].find { |f| File.exists?(f) }
```
haus: this could be `Dir["#{repo}/VERSION", "#{repo}/bin/VERSION"].first`. Not sure if that's any cleaner.

Iristyle: :+1: thanks

Iristyle:
As soon as I can get a new build of cfacter off of Jenkins that includes a `VERSION` file generated via https://github.com/puppetlabs/cfacter/commit/4a9fdd1 ... then I can do another run of `puppet_for_the_win` locally to verify this works.

Iristyle:
Verified as correct with the following YAML

```yaml
---
:repos:
  puppet:
    :ref: origin/master
    :repo: git://github.com/puppetlabs/puppet.git
  facter:
    :ref: 2.4.0
    :repo: git://github.com/puppetlabs/facter.git
  cfacter:
    :archive: 'cfacter-0.3.0-72-g4a9fdd1-x64.zip'
    :path: 'http://builds.puppetlabs.lan/cfacter/4a9fdd197cbd9a31a466e3688cb32935032f73aa/artifacts/windows/'
  hiera:
    :ref: 1.3.4
    :repo: git://github.com/puppetlabs/hiera.git
  sys:
    :ref: origin/2.1.x-x64
    :repo: git://github.com/puppetlabs/puppet-win32-ruby.git
  mcollective:
    :ref: 2.8.0
    :repo: git://github.com/puppetlabs/marionette-collective.git
```

Note that the artifact `cfacter-0.3.0-72-g4a9fdd1-x64.zip` includes `bin/VERSION` that is written as ASCII based on https://github.com/puppetlabs/cfacter/commit/4a9fdd197cbd9a31a466e3688cb32935032f73aa

```
> type .\stagedir\misc\versions.txt

cfacter 0.3.0-72-g4a9fdd1
facter 2.4.0-0-g8ff4fa5
hiera 1.3.4-0-g7d21886
mcollective 2.8.0-0-gd3c6dce
puppet 3.7.4-1381-gb90b1ec
sys 2.1.5.0-x64-0-gb313f5b
```

tasks/windows/windows.rake
```diff
@@ -223,7 +223,14 @@ namespace :windows do
     version_tracking_file = 'stagedir/misc/versions.txt'
     content = ""
     FileList["downloads/*"].each do |repo|
-      content += "#{File.basename(repo)} #{describe repo}\n"
+      is_git = File.exists?("#{repo}/.git")
+      if is_git
+        ver = describe(repo)
+      else
+        ver_path = Dir["#{repo}/VERSION", "#{repo}/bin/VERSION"].first
+        ver = File.read(ver_path).strip
```
ferventcoder: What if ver_path is nil? Other than that I'm +1

ferventcoder: Also why `ver` versus a maybe something a bit more descriptive?

Iristyle: I think the script should fail with an error when `ver_path` is nil.

ferventcoder: I would prefer if it failed, it would be a gracious failure - so that the person that gets the failure knows what their next steps are to take. It won't be obvious - something like this would happen far and few between. It would be better to tell them that "for ___ application they need to have the build of it include a version file since it is not a git repo." or something like that.

ferventcoder: @haus @melissa ^^ What path is best to take here? I'll leave the decision up to you on that.

tasks/windows/windows.rake
```diff
@@ -223,7 +223,14 @@ namespace :windows do
     version_tracking_file = 'stagedir/misc/versions.txt'
     content = ""
     FileList["downloads/*"].each do |repo|
-      content += "#{File.basename(repo)} #{describe repo}\n"
+      is_git = File.exists?("#{repo}/.git")
+      if is_git
+        ver = describe(repo)
+      else
+        ver_path = Dir["#{repo}/VERSION", "#{repo}/bin/VERSION"].first
```
joshcooper: does `Dir[]` exist in ruby 1.8? do we care?

haus: it does: http://ruby-doc.org/core-1.8.7/Dir.html#method-c-5B-5D

Iristyle: `Dir[]` does, but `Dir.exists?` does not.  I did verify this code ran on 1.8

tasks/windows/windows.rake
haus: adding some simple `if ver_path...else warn && exit` seems sane to me.

Iristyle:
@ferventcoder this is ready to merge

haus:
:+1:

PR #99, opened 2015-02-24
(QENG-1892) Configure git clone with less memory

Iristyle:
- Git clones of large repositories are failing in more resource
   constrained environments (such as the VMPooler instances that the
   new AIO pipeline is using).  This has resulted in errors like

```
     Resolving deltas:  46% (3744/8085)
     error: index-pack died of signal 11
     fatal: index-pack failed
     rake aborted!
```
   To address this issue, configure Git prior to cloning to adjust two
   key settings:

####   pack.windowMemory
   The maximum size of memory that is consumed by each thread in
   git-pack-objects[1] for pack window memory when no limit is given on
   the command line. The value can be suffixed with "k", "m", or "g".
   When left unconfigured (or set explicitly to 0), there will be no limit.

####   pack.packSizeLimit
   The maximum size of a pack. This setting only affects packing to a
   file when repacking, i.e. the git:// protocol is unaffected. It can
   be overridden by the --max-pack-size option of git-repack[1]. The
   minimum size allowed is limited to 1 MiB. The default is unlimited.
   Common unit suffixes of k, m, or g are supported.

   Some information discovered in SO thread:
   http://stackoverflow.com/questions/4826639/#4829883

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -146,7 +146,14 @@ namespace :windows do
               sh "rm #{config[:archive]}"
             end
           else
-            sh "git clone #{config[:repo]} #{name}"
+            FileUtils.mkdir(name)
+            Dir.chdir "#{TOPDIR}/downloads/#{name}" do
+              sh "git init"
+              sh "git remote add origin #{config[:repo]}"
+              sh "git config pack.windowMemory 10m"
+              sh "git config pack.packSizeLimit 20m"
+              sh "git fetch"
+            end
           end
         end
       end


Iristyle:
@joshcooper @justinstoller @haus

joshcooper:
@Iristyle if you haven't already could you grab a 2012r2 vm from the pool, and execute those commands to clone puppet-win32-ruby?

Iristyle:
@joshcooper I have executed the commands in a local environment, but not yet a VMPooler VM:

* By building the whole MSI through the  `puppet_for_the_win` using `bundle exec rake windows:build CONFIG=foss-4.0-x64.yaml`
* In Windows 2008R2
* With 2GB memory
* Git version 1.9.5

Give me a few minutes to grab a Vmpooler vm, install Choc, install Git 1.9.5 ... and clone `puppet-win32-ruby` a few times.

Iristyle:
Interestingly, I'm not able to repro on the 2012r2 VMPooler VM (which has 4GB of memory btw)... I think I would need to simulate more of a `real build` or otherwise induce a low memory situation.  

However, I did not that this PR is incorrect as it stands.  The `git config` is actually modifying the parent `puppet_for_the_win` repo and not the fresh clones.

Iristyle:
PR has been reworked to ensure settings are modified on the appropriate local repo prior to cloning.

tasks/windows/windows.rake
```diff
@@ -146,7 +146,14 @@ namespace :windows do
               sh "rm #{config[:archive]}"
             end
           else
-            sh "git clone #{config[:repo]} #{name}"
+            sh "md #{name}"
```
haus: Any reason this can be `FileUtils.mkdir` or `FileUtils.mkdir_p`?

haus:
Aside from my fileutils comment, this looks reasonable. Are these settings something we would want to apply globally to the jenkins git config?

Iristyle:
@haus you could make a case for applying these settings globally... though we don't know yet for 100% certain that these are the extent of the changes.

Also, that would be a move from explicit behavior (that's easy to see in the log output) to implicit / magical behavior... which I would be less in favor of.

haus:
@Iristyle yea, i can see the case for that. i was more concerned with needing to apply the same git changes in a bunch of places, but we can burn that bridge when we see it.

haus:
:+1:

haus:
:+1:

ferventcoder:
:+1:

PR #100, opened 2015-02-25
(RE-4056) Log environment variables / tools

Iristyle:
- Based on how the new AIO pipeline is invoking the Rake tasks, it was
   determined that builds were using two different sets of tools.  This
   commit simply logs what the current environment variables are (which
   is useful for diagnosing PATH issues), and it also outputs tool
   versions for the commonly used tools in the Rake tasks.

   Unfortunately, candle, light and heat don't have --version or similar
   switches.

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -307,6 +307,30 @@ namespace :windows do
   # High Level Tasks.  Other tasks will add themselves to these tasks
   # dependencies.
 
+  def safe_sh(cmd_line)
+    begin
+      sh cmd_line
+    rescue
+      puts "Failed to execute command line #{cmd_line}"
+    end
+  end
+
+  desc "Describe environment variables and tooling"
+  task :log_environment do |t|
+    puts "Current environment variables:"
+    ENV.each { |k, v| puts "#{k}: #{v}" }
+
+    puts "Tooling versions:\n"
+
+    safe_sh("git --version")
+    safe_sh("ruby --version")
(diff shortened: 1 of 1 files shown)


joshcooper:
You could do `candle -help`, etc which includes the version info along with usage... not sure if we want to do that.

Iristyle:
Add `candle -help` @joshcooper  - thanks.

tasks/windows/windows.rake
```diff
@@ -307,6 +307,29 @@ namespace :windows do
...
+    ENV.each { |k, v| puts "#{k}: #{v}" }
+
+    puts "Tooling versions:\n"
+
+    safe_version("git --version")
+    safe_version("ruby --version")
+    safe_version("rake --version")
+    safe_version("curl --version")
+    safe_version("candle -help")
+    safe_version("unzip -v")
```
ferventcoder: I think in addition this could be 7zip?

Iristyle: Yeah, I could do that too... `7za` is another funny one without any `-version` switch or similar.

melissa: We were talking about how necessary it was to know the version of `7za`, even of `unzip`. It might be more interesting to see if they are even available on the system?

Iristyle:
Merged based on pairing with @melissa / comments from @joshcooper / @ferventcoder

joshcooper:
:+1:

ferventcoder:
:+1:

PR #101, opened 2015-03-03
(PUP-4067) Add package manifest and module dirs

melissa:
In order to support the puppet-agent run in a masterless setting, we
must provide these directories via the puppet agent installation, rather
than the puppet master installation.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -357,6 +357,27 @@
                 <CreateFolder />
               </Component>
             </Directory>
+            <Directory Id="EnvironmentsDir" Name="environments">
+              <Component Id="EnvironmentsDir" Permanent="yes" Guid="6a6a390a-8783-40b6-b49d-9e2b368e069d">
+                <CreateFolder />
+              </Component>
+              <Directory Id="ProductionDir" Name="production">
+                <Component Id="ProductionDir" Permanent="yes" Guid="b902efcd-756e-4340-8f13-5a79d46b93c7">
+                  <CreateFolder />
+                  <File Id="EnvironmentConf" KeyPath="yes" Source="stagedir\puppet\conf\environment.conf" />
+                </Component>
+                <Directory Id="ManifestsDir" Name="manifests">
+                  <Component Id="ManifestsDir" Permanent="yes" Guid="5484fadf-6d5e-4053-83fc-ea68918d6f2e">
+                    <CreateFolder />
+                  </Component>
+                </Directory>
+                <Directory Id="ModulesDir" Name="modules">
+                  <Component Id="ModulesDir" Permanent="yes" Guid="cbe9cc8d-3a38-4c4c-9f16-39e42093481a">
+                    <CreateFolder />
+                  </Component>
+                </Directory>
+              </Directory>
+            </Directory>
           </Directory>
         </Directory>
       </Directory>
@@ -749,6 +770,10 @@
       <ComponentRef Id="PuppetVarDir" />
       <ComponentRef Id="CodeDir" />
       <ComponentRef Id="HieraDataDir" />
+      <ComponentRef Id="EnvironmentsDir" />
+      <ComponentRef Id="ProductionDir" />
+      <ComponentRef Id="ManifestsDir" />
+      <ComponentRef Id="ModulesDir" />
       <ComponentRef Id="PuppetShortcuts" />
       <ComponentRef Id="PuppetDocumentationShortcuts" />
 


wix/puppet.wxs
```diff
@@ -357,6 +357,26 @@
...
+                  <CreateFolder />
+                </Component>
+                <Directory Id="ManifestsDir" Name="manifests">
+                  <Component Id="ManifestsDir" Permanent="yes" Guid="">
+                    <CreateFolder />
+                  </Component>
+                </Directory>
+                <Directory Id="ModulesDir" Name="modules">
+                  <Component Id="ModulesDir" Permanent="yes" Guid="">
+                    <CreateFolder />
```
joshcooper: I think these need GUIDs, the wix docs have a link about how to create these: http://wixtoolset.org/documentation/manual/v3/howtos/general/generate_guids.html

melissa: Yes, definitely. I hadn't gotten to that part yet.

melissa: Thanks for pulling up that link for me! Guids have been added

melissa:
@joshcooper this should be ready now. I've built it, and it looks like it's doing the correct thing.

melissa:
screen shot 2015-03-03 at 3 39 45 pm

haus:
:+1:

wix/puppet.wxs
```diff
@@ -749,6 +769,10 @@
       <ComponentRef Id="PuppetVarDir" />
       <ComponentRef Id="CodeDir" />
       <ComponentRef Id="HieraDataDir" />
+      <ComponentRef Id="EnvironmentsDir" />
+      <ComponentRef Id="ProductionDir" />
+      <ComponentRef Id="ManifestsDir" />
+      <ComponentRef Id="ModulesDir" />
```
joshcooper: It surprises me that we need to add more than just the parent `EnvironmentsDir`, but it's been awhile since I've used wix, and it's working, so :+1:

melissa:
We might not have to, but I think since each directory is also a component (or has a component id), it needs to be added.

wix/puppet.wxs
```diff
@@ -357,6 +357,26 @@
                 <CreateFolder />
               </Component>
             </Directory>
+            <Directory Id="EnvironmentsDir" Name="environments">
+              <Component Id="EnvironmentsDir" Permanent="yes" Guid="6a6a390a-8783-40b6-b49d-9e2b368e069d">
```
joshcooper: Unfortunately I think these need to be uppercase: 

> The lowercase letters in these GUIDs must be changed to uppercase to be used as a product code or package code.

https://msdn.microsoft.com/en-us/library/aa370568%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396

It may also be that only earlier versions of windows installer (in 2003) had this problem

melissa:
I have yes to test this with Haus's addition that adds in `environment.conf`, but it is no longer blocked on anything

Iristyle:
Do you want to get rid of the PR to the PR in a rebase @melissa ?

melissa:
Hmm, so, this isn't installing `environment.conf`. I'm going to dig in a little more and see if I can figure out why

wix/puppet.wxs
```diff
@@ -357,6 +357,27 @@
...
+                <CreateFolder />
+                <File Id="EnvironmentConf" KeyPath="yes" Source="stagedir\puppet\conf\environment.conf" />
```
melissa: Also, @haus do we want this file in `environments/production`? Right now it looks like it's going to drop it into `environments`

haus: everything!

melissa:
Okay, this has been verified now. I needed to pull out the mcollective version updates to get it working, so I was lazy and just included that work in this PR. Let me know if you want me to pull it out into a different PR

Iristyle:
@melissa #103 has been merged if you want to do a quick update on this PR?

melissa:
@Iristyle updated

ferventcoder:
I'm not sure what the cut up veggies are all about... but :+1:

wix/puppet.wxs
ferventcoder: In this set the unnecessary components are `EnvironmentsDir` and `ProductionDir` - since there are components in subfolders below them  (ManifestsDir and ModulesDir), they would have created the folder hierarchy for `environments/production` appropriately.

ferventcoder: No worries though, it all works.

melissa:
@ferventcoder They wanted me to "squash" things. We had both a summer squash and a winter squash to help with that message

wix/puppet.wxs
melissa: We can always pull those out if it's annoying enough. That's good information to have though

ferventcoder:
Well that's interesting :)

PR #102, opened 2015-03-04
(MCO-597) Ensure mco.bat

ferventcoder:
Mco should be available as a command from the command line so ensure
that mco.bat is copied to the bin directory to be picked up as a
fragment.

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -178,6 +178,7 @@ namespace :windows do
 
     # Only copy the .bat files into place
     cp_p(FileList["conf/windows/stage/bin/*.bat"], "stagedir/bin/")
+    FileUtils.cp('downloads/mcollective/ext/windows/mco.bat', 'stagedir/bin/mco.bat') if File.exists?('downloads/mcollective/ext/windows/mco.bat')
   end
 
   task :misc => 'stagedir' do


tasks/windows/windows.rake
```diff
@@ -178,6 +178,7 @@ namespace :windows do
 
     # Only copy the .bat files into place
     cp_p(FileList["conf/windows/stage/bin/*.bat"], "stagedir/bin/")
+    FileUtils.cp('downloads/mcollective/ext/windows/mco.bat', 'stagedir/bin/mco.bat') if File.exists?('downloads/mcollective/ext/windows/mco.bat')
```
Iristyle: Shouldn't this be an error if it doesn't exist?

ferventcoder: Not necessarily. What happens if you don't include mco as part of the yaml? This isn't a normal situation for us, but could be for the community. This also follows the pattern below it.

ferventcoder: @haus @melissa thoughts on the above?

haus:
@ferventcoder not really. we don't enforce mco inclusion today, but we also don't enforce inclusion of any projects, so I don't feel very strongly about this.

tasks/windows/windows.rake
joshcooper: I would rather fail hard. The file is part of the AIO specification and if it's not there, something is wrong.

Iristyle: I'm also :+1: to failing hard here as well for the same reason.  I'll adjust the PR so that we can get this merged today.

ferventcoder: Should we update the other MCO bits that check to see if it is there?

PR #105, opened 2015-03-16
(maint)(packaging) Cherry-pick fixes for mcollective 2.8 release to stable

shrug:
This is a cherry-pick of 2 commits with fixes related to the Marionette Collective 2.8.x release to the stable branch. One related to moving the plugins into the lib dir, and the other updating the version handling for the Marionette Collective packaging

### DIFF
--- a/conf/windows/stage/bin/environment.bat.erb
+++ b/conf/windows/stage/bin/environment.bat.erb
@@ -21,7 +21,6 @@ SET FACTER_DIR=%PL_BASEDIR%\facter
 SET HIERA_DIR=%PL_BASEDIR%\hiera
 <%- if ENV["BRANDING"] == "enterprise" -%>
 SET MCOLLECTIVE_DIR=%PL_BASEDIR%\mcollective
-SET MCOLLECTIVE_PLUGIN_DIR=%PL_BASEDIR%\mcollective_plugins
 <% end -%>
 
 <%- if ENV["BRANDING"] == "enterprise" -%>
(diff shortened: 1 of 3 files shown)


ferventcoder:
So are we using stable now?

shrug:
Sorry, more context: The compose is currently failing for PE 3.8.x builds, and updating the job to the sha that includes these fixes brings in other, unwanted (for 3.8) changes.

shrug:
I verified that if you update to the sha  that includes the MCO versioning fixes, the compose fails due to missing cfacter. So our options (as I see it) are move the build to stable with the fixes required, make a new branch, or revert the MCO 2.8 update

melissa:
So, moving the mcollective plugin location here seems like a big change. Is that something we're okay with releasing in PE 3.8? If so, I'm :+1: on this

haus:
It should be transparent to users. Any plugins delivered via module will still go to the old location which will still be in the libdir in server.cfg.

haus:
@shrug if we're using this for PE 3.8, I don't think we can burn the version swap wholesale? Puppet still needs it to swap in the PE version correctly (if that code still works).

shrug:
OK, I'm cargo culting that a bit, I have not looked into that enough to really understand what's going on. I'll dig in a bit.

shrug:
Updated, I reverted the cherry-pick that removed the entire version task, and just stripped out the mcollective bits.

haus:
@shrug i *think* you missed clipping an end out of the version section (the close to the mco, pup.each block). also tabbing in that section is off.

shrug:
Good eye, thanks.

haus:
:+1: this looks amazing.

PR #106, opened 2015-04-02
Revert "(maint) Ensure there are no spaces in the msi name"

melissa:
This reverts commit f6584df2d8f1a9318ae4a8448d9d2a64237743cb.

### DIFF
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -120,7 +120,7 @@ namespace :windows do
   CONFIG = YAML.load_file(ENV["config"] || "config.yaml")
   APPS = CONFIG[:repos]
   ENV['ARCH'] = ENV['ARCH'] || 'x86'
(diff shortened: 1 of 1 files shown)


ferventcoder:
:-1: Don't see what this hurts.

ferventcoder:
I don't have a problem guarding against user input.

haus:
@melissa it's unclear from the commit message why this should be reverted

Iristyle:
I think the problem was that it covered problems in how build parameters were passed through the pipeline, so I'm :+1: to merging this.

melissa:
@Iristyle @ferventcoder commit message updated. I'm fine with either leaving this in, or pulling it out. Whatever seems to make more sense.

melissa:
@ferventcoder @Iristyle any advice on how to move forward with this? I'd like to resolve the ticket.

ferventcoder:
LGTM. I'm +1 to merging

PR #107, opened 2015-04-09
(MCO-625) Add mco.bat / required environment vars

ferventcoder:
Drop the use of the mco.bat from the marionette-collective repo
and instead prefer the use of an mco.bat included in PFW. This is
similar to how other bat files are included, e.g. puppet.bat and
facter.bat.

Also inlucde an environment variable for mcollectived.

This reverts changes introduced with https://github.com/puppetlabs/puppet_for_the_win/commit/07ea56edf9cbf96ffb920e35d9629d79cab339e5

### DIFF
--- a/conf/windows/stage/bin/mco.bat
+++ b/conf/windows/stage/bin/mco.bat
@@ -1,0 +1,6 @@
+@echo off
+SETLOCAL
+
+call "%~dp0environment.bat" %0 %*
+
+ruby -S -- "%SCRIPT_NAME%" %*
--- a/tasks/windows/windows.rake
+++ b/tasks/windows/windows.rake
@@ -178,7 +178,6 @@ namespace :windows do
 
     # Only copy the .bat files into place
     cp_p(FileList["conf/windows/stage/bin/*.bat"], "stagedir/bin/")
-    FileUtils.cp('downloads/mcollective/ext/windows/mco.bat', 'stagedir/bin/mco.bat')
   end
 
   task :misc => 'stagedir' do


conf/windows/stage/bin/environment.bat
```diff
@@ -22,6 +22,13 @@ SET CFACTER_DIR=%PL_BASEDIR%\cfacter
 SET HIERA_DIR=%PL_BASEDIR%\hiera
 SET MCOLLECTIVE_DIR=%PL_BASEDIR%\mcollective
 
+REM Set mco.bat file environment variables
+SET DATA_DIR=%ALLUSERSPROFILE%\PuppetLabs
+REM This could also be ~/.mcollective/etc/server.cfg if the user is not running as admin
```
ferventcoder: This is new, we've never needed to set the path to a config from here before. We may not need it.

ferventcoder: I have not dived in too far yet

ferventcoder:
https://github.com/puppetlabs/marionette-collective/pull/306

conf/windows/stage/bin/environment.bat
ferventcoder: This is not necessary.

conf/windows/stage/bin/environment.bat
```diff
@@ -21,6 +21,7 @@ SET FACTER_DIR=%PL_BASEDIR%\facter
 SET CFACTER_DIR=%PL_BASEDIR%\cfacter
 SET HIERA_DIR=%PL_BASEDIR%\hiera
 SET MCOLLECTIVE_DIR=%PL_BASEDIR%\mcollective
+SET MCOLLECTIVED=%MCOLLECTIVE_DIR%\bin\mcollectived
```
joshcooper: we're setting this env var, but it doesn't look like it's used anywhere, unless there's some magic?

ferventcoder: This was brought over since it was the one we were not quite sure about with respect to mco and how it may use that.

ferventcoder: But it would likely be picked up as an env var in the app. If that is not the case, I'm happy to remove it.

ferventcoder: @richardc I'm not sure what we decided here. I'm guessing we probably don't need this?

richardc: My guess is we don't need it.  I expect it was made irrelevant with 088c568dc3

ferventcoder:
This is ready again @joshcooper @richardc

richardc:
Seems sane to me :+1:

conf/windows/stage/bin/environment.bat
ferventcoder: @Iristyle here's the original items.

PR #110, opened 2015-04-25
(maint) Remove accidental Facter 'dir' fact

Iristyle:
- Facter will look for environment variables that begin with FACTER_
   and will automatically create facts out of them.  Inside of the
   environment.bat file called prior to the Puppet command line
   wrappers (puppet.bat, facter.bat, hiera.bat, etc) the temporary
   variable FACTER_DIR was used to set up the PATH.

   This had the interesting side effect of producing the DIR fact,
   which was totally unintended and unspecified.

   Simply rename this temp variable so that this doesn't happen.


MikaelSmith:
This will be picked up by stable builds of puppet-agent.

PR #114, opened 2015-06-08
(PUP-4631) Introduce Windows 2003 deprecation msg

Iristyle:
- During installation, warn users about installing on Windows 2003.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -226,7 +226,7 @@
                       Start="demand"
                       Type="ownProcess"
                       ErrorControl="normal"
-                      Vital="yes" 
+                      Vital="yes"
                       Arguments="-I&quot;[INSTALLDIR]puppet\lib;[INSTALLDIR]facter\lib;[INSTALLDIR]hiera\lib;[INSTALLDIR]mcollective\lib&quot; -rubygems &quot;[INSTALLDIR]mcollective\bin\mcollectived&quot; --daemonize"
                       />
                     <ServiceControl Id="MCOStartService" Stop="both" Remove="uninstall" Name="mcollective" Wait="yes" />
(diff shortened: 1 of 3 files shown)


wix/puppet.wxs
```diff
@@ -702,6 +702,15 @@
       </Custom>
     </InstallUISequence>
     <InstallExecuteSequence>
+      <!-- Unsupported Windows version check -->
+      <Custom Action="DeprecatedWindowsWarning" After="InstallInitialize">
+        <![CDATA[VersionNT64 < 600]]>
```
ferventcoder: I'm guessing these are when conditions are true as compared to when the check fails, such as with https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L61.

wix/puppet.wxs
```diff
@@ -702,6 +702,15 @@
...
+      </Custom>
+      <CustomAction Id="DeprecatedWindowsWarning" Script="vbscript">
+        <![CDATA[
+        session.Message("Windows Server 2003 and 2003R2 have been officially end-of-lifed by Microsoft as of July 15, 2015.  This $(var.OurProductName) package is deprecated for use on this platform.&#xD;&#xA;&#xD;&#xA;Future $(var.OurProductName) packages will no longer be supported on these platforms.  &#xD;&#xA;&#xD;&#xA;Please plan accordingly and see http://links.puppetlabs.com/deprecate-windows-2003 for additional information.")
```
ferventcoder: Why vbscript?

Iristyle: Old code.. I already changed to `jscript`.  But it doesn't really matter given it's basically identical code.

wix/puppet.wxs
Iristyle: I believe so (though still testing and having a tough time of getting it to work).

I think the basis for comparison you would want though is https://github.com/puppetlabs/puppet_for_the_win/blob/master/wix/puppet.wxs#L752

The check you linked to is a different type and blocks the installation.... AFAIK, we only want to add a warning.  It may turn out that we want to add a customized dialog instead (though that's a bit more complex and would be great to avoid if I can get this `Session.Message` business working)

ferventcoder: `VersionNT` is what we should look at instead of `VersionNT64`

wix/puppet.wxs
```diff
@@ -657,6 +668,10 @@
          restore the properties set by the command line overriding the recalled
          value from the registry -->
     <InstallUISequence>
+      <!-- Unsupported Windows version check -->
+      <Custom Action="DeprecatedWindowsWarning" After="LaunchConditions">
```
Iristyle: `LaunchConditions` stage was selected based off the ordering published in [this AdvancedInstaller Guide](http://www.advancedinstaller.com/user-guide/standard-actions.html)

Iristyle:
FYI - we evaluated a couple of options here ... including adding a custom dialog.  See:

* http://stackoverflow.com/questions/1810910/how-to-display-a-non-blocking-warning-for-the-operating-system-in-wix
* http://stackoverflow.com/questions/6652211/wix-how-to-warn-user-instead-of-terminating-installation-using-properties

Given scheduling, we tried to minimize changes to the MSI, so opted for a simpler message box approach.

ferventcoder:
@Iristyle Is this ready to go?

PR #116, opened 2015-06-11
(PUP-4631) Windows 2003 Deprecation Dialog

Iristyle:
- Introduce a new dialog into the UI sequence that shows after the
   Welcome Dialog on Windows 2003 only.  It displays a link to more
   information on the upgrade path for Puppet from Windows 2003.
 - The dialog is dynamically injected into *only* the new install and
   patch UI sequences and not maintenance, based on the VersionNT
   property < 600.
 - We are able to use only the VersionNT (and not VersionNT64) property
   in this situation, because the installation is already blocked for
   the x64 installer on Windows 2003.
 - Note that there is nothing written to the installation log warning
   about 2003.  It appeared that a custom action was the only solution,
   and we're trying to avoid custom actions.
 - The previous attempt at providing this functionality in 32e1aeb,
   though simpler in code change terms, was actually deemed to be
   unsuitable.  The custom VBScript action had issues in a Cygwin
   environment, and after futher research, it was discovered that
   custom VBScript / JScript actions should never be used, because
   amongst other reasons, over-eager AV software may terminate the
   scripts altogether causing interesting failures.  For more
   information, see:

   http://blogs.msdn.com/b/robmen/archive/2004/05/20/136530.aspx

### DIFF
--- a/wix/ui/PuppetWelcomeDlg.wxs
+++ b/wix/ui/PuppetWelcomeDlg.wxs
@@ -1,13 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
     Copyright (c) Microsoft Corporation.  All rights reserved.
-    
+
     The use and distribution terms for this software are covered by the
     Common Public License 1.0 (http://opensource.org/licenses/cpl1.0.php)
     which can be found in the file CPL.TXT at the root of this distribution.
     By using this software in any fashion, you are agreeing to be bound by
     the terms of this license.
-    
+
     You must not remove this notice, or any other, from this software.
 -->
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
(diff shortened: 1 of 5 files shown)


wix/ui/PuppetWelcomeDlg.wxs
```diff
@@ -21,7 +21,9 @@
                     <Publish Event="SpawnDialog" Value="CancelDlg">1</Publish>
                 </Control>
                 <Control Id="Bitmap" Type="Bitmap" X="0" Y="0" Width="370" Height="234" TabSkip="no" Text="!(loc.WelcomeDlgBitmap)" />
-                <Control Id="Back" Type="PushButton" X="180" Y="243" Width="56" Height="17" Disabled="yes" Text="!(loc.WixUIBack)" />
+                <Control Id="Back" Type="PushButton" X="180" Y="243" Width="56" Height="17" Disabled="yes" Text="!(loc.WixUIBack)">
+                    <Condition Action="hide">1</Condition>
+                </Control>
```
ferventcoder: Guess I never realized we had a back button on the welcome dialog by default.

PR #118, opened 2015-09-04
(RE-5539) Set SSL environment variables for windows processes

melissa:
We want to ensure OpenSSL has secure settings, and the defaults are not
guarenteed to give us that. This commit updates SSL_CERT_FILE and
SSL_CERT_DIR to make sure we have safe locations set for both.

### DIFF
--- a/conf/windows/stage/bin/environment.bat
+++ b/conf/windows/stage/bin/environment.bat
@@ -33,3 +33,7 @@ SET RUBYLIB=%RUBYLIB:\=/%
 REM Enable rubygems support
 SET RUBYOPT=rubygems
 REM Now return to the caller.
+
+REM Set SSL variables to ensure trusted locations are used
+SET SSL_CERT_FILE=%WINDIR%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%WINDIR%\system32\ssl\certs


joshcooper:
LGTM, minor nit, could you update the subject line in the commit itself?

Iristyle:
:+1: aside from `SLL` in the commit message that @joshcooper mentioned.

melissa:
done :)

haus:
https://en.wikipedia.org/wiki/Windir

conf/windows/stage/bin/environment.bat
```diff
@@ -33,3 +33,7 @@ SET RUBYLIB=%RUBYLIB:\=/%
 REM Enable rubygems support
 SET RUBYOPT=rubygems
 REM Now return to the caller.
+
+REM Set SSL variables to ensure trusted locations are used
+SET SSL_CERT_FILE=%WINDIR%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%WINDIR%\system32\ssl\certs
```
ferventcoder: I'm late on this, but I think I'd prefer `%systemroot%` over `%windir%`. One is built-in and the other just uses it by default. http://superuser.com/questions/638321/what-is-difference-between-windir-and-systemroot

ferventcoder: Since we are talking about safer defaults, `SystemRoot` is read-only and cannot be changed system-wide, where `WinDir` is included for compatibility with older Windows systems programs and by default uses `SystemRoot`. 

`WinDir` has three attack vectors:
 * Change the environment variable systemwide.
 * Overload any environment variable over the MAX_PATH and none of them (aside from built-in) work.
 * Change the environment variable in-process.

`SystemRoot` has this attack vector:
 * Change the environment variable in-process.


Exploiting "Changing the variable in-process" would likely require tricking the user context to run a specially crafted command to change the variable. A sub-process would not be able to change it.

ferventcoder: A sub-process would not be able to change the in-proc env var. It is much harder to exploit versus the other two susceptible attack vectors for WinDir.

ferventcoder: To be clear I don't consider overloading an environment variable over `MAX_PATH` as something that would be dangerous, it would just cause things to stop working. It's the ability to exploit pointing to somewhere else system wide that I see as a possible concern.

Iristyle: Interesting notes - I didn't know that.

What about situations where the Windows install dir is `windows.000` or similar rather than `%SystemRoot%\Wijndows` though?

ferventcoder: @Iristyle `%SystemRoot%`, not `%SystemDrive%`. If you look at `%WinDir%`, you will notice it is set to `%SystemRoot%`, which means `C:\Windows` most of the time.

ferventcoder: Not all of the time though is the above true.

ferventcoder: image

Iristyle: My mistake @ferventcoder - thanks for pointing out.

I agree that `%SystemRoot%` seems like a better choice.

melissa: https://github.com/puppetlabs/puppet_for_the_win/pull/120

PR #120, opened 2015-09-08
(RE-5539) Update SSL settings to secure default 

melissa:
Prior to this change, we were using %WINDIR% to define where our SSL
variables were set. This, however, is an insecure way of doing it.
%WINDIR% has three attack vectors associated with it, whereas
%SYSTEMROOT% only has one. The one attack vector for %SYSTEMROOT% is by
far more difficult to exploit. Because of this, we want to default  to
%SYSTEMROOT% rather than %WINDIR%. In a generic setup, %WINDIE% defaults
to %SYSTEMROOT%.

### DIFF
--- a/conf/windows/stage/bin/environment.bat
+++ b/conf/windows/stage/bin/environment.bat
@@ -35,5 +35,5 @@ SET RUBYOPT=rubygems
 REM Now return to the caller.
 
 REM Set SSL variables to ensure trusted locations are used
-SET SSL_CERT_FILE=%WINDIR%\system32\ssl\cert.pem
-SET SSL_CERT_DIR=%WINDIR%\system32\ssl\certs
+SET SSL_CERT_FILE=%SystemRoot%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%SystemRoot%\system32\ssl\certs


conf/windows/stage/bin/environment.bat
```diff
@@ -35,5 +35,5 @@ SET RUBYOPT=rubygems
 REM Now return to the caller.
 
 REM Set SSL variables to ensure trusted locations are used
-SET SSL_CERT_FILE=%WINDIR%\system32\ssl\cert.pem
-SET SSL_CERT_DIR=%WINDIR%\system32\ssl\certs
+SET SSL_CERT_FILE=%SystemRoot%\system32\ssl\cert.pem
+SET SSL_CERT_DIR=%SystemRoot%\system32\ssl\certs
```
joshcooper: Sadly `%SystemRoot%` does not include the drive letter like %WINDIR% so this needs to be `%SystemDrive%%SystemRoot%` https://en.wikipedia.org/wiki/Environment_variable#Default_values

ferventcoder: @joshcooper ??

If you will notice from this image, %WINDIR% is set to %SYSTEMROOT%, so they should evaluate EXACTLY the same.


image

joshcooper: I am imaging things... I thought `%SystemRoot%` didn't include the `%SystemDrive%`, but it clearly does.

melissa:
Yes, this should have been targeted at master

melissa:
https://github.com/puppetlabs/puppet_for_the_win/pull/121

conf/windows/stage/bin/environment.bat
ferventcoder: Haha, no worries mate! :)

PR #123, opened 2015-09-30
(RE-5642) Adding pxpagent - first pass

jcoconnor:
(RE-5642) First pass at adding pxpagent to the build.

### DIFF
--- a/wix/PuppetForTheWin.wixproj
+++ b/wix/PuppetForTheWin.wixproj
@@ -26,6 +26,7 @@
     <Compile Include="fragments\facter.wxs" />
     <Compile Include="fragments\hiera.wxs" />
     <Compile Include="fragments\misc.wxs" />
+    <Compile Include="fragments\pxp-agent.wxs" />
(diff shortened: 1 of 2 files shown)


ferventcoder:
This probably also needs the updates to a sample yaml file to see how this is intended to be added.

ferventcoder:
I'm guessing it is just a zip file for the most part that would be added that would download the agent and grab the files it uses to produce the wxs file (heat).

jcoconnor:
Amended comment in commit to add ticket number.

@ferventcoder Can you elaborate on what you are looking for with a sample yaml file ?

This change is associated with PR https://github.com/puppetlabs/puppet-agent/pull/311 which generates a pxp-agent.zip containing the pxp-agent assets for staging.

ferventcoder:
@jcoconnor sure thing - take a look at all of the yaml files in the top level directory of this repo (https://github.com/puppetlabs/puppet_for_the_win). Edit one of them to add the zip archive details

PR #124, opened 2015-10-08
(RE-5640) Add nssm to the MSI manifest

jcoconnor:
Adding the nssm package to the MSI manifest so that is installed under C:\Program Files
\Puppet Labs\nssm

Note https://github.com/puppetlabs/puppet-agent/pull/327 needs to be merged before this PR.


Iristyle:
Just as a follow up, I think we want to keep {{nssm.exe}} with the other service related files as noted in the ticket.

See the current rake task for info:
https://github.com/puppetlabs/puppet_for_the_win/blob/master/tasks/windows/windows.rake#L188-L192

jcoconnor:
Agreed @Iristyle - am revising this to just copy the correct (32/64) nssm.exe to stagedir, depending on the arch setting of the build.

jcoconnor:
@Iristyle Updated with revised nssm.exe code.
Its not actually getting to the installed service directory, so going to look at fixing that for tomorrow.
Will also tidy up the commits then.

tasks/windows/windows.rake
```diff
@@ -189,6 +189,13 @@ namespace :windows do
     mkdir_p("stagedir/service")
     FileUtils.cp_r(FileList['downloads/puppet/ext/windows/service/*'], 'stagedir/service')
     FileUtils.cp('downloads/mcollective/ext/windows/daemon.bat', 'stagedir/service/mco_daemon.bat') if File.exists?('downloads/mcollective/ext/windows/daemon.bat')
+
+    # setup nssm - note - we are clobbering the downloads/nssm directory deliberately to stop the ":stage" task from copying nssm
+    # to stagedir and creating an (unused) wxs fragment file.
+    nssmwindir = ENV['ARCH'] == 'x64' ? 'win64' : 'win32'
+    FileUtils.cp("downloads/nssm/#{nssmwindir}/nssm.exe", 'stagedir/service/nssm.exe')
```
ferventcoder: Aren't there more files necessary here?

jcoconnor: @ferventcoder Do you mean .bat files or additional ?
nssm.exe is actually standalone afaik.

ferventcoder: I meant other files nssm.exe may need to have access to (dlls or others).

PR #126, opened 2015-10-15
(RE-5641) Adding PXP-Agent as service under NSSM

jcoconnor:
Use NSSM to daemonize pxp-agent as a Windows service.
This copies the MCO approach where the MSI does the initial
install and uses a PE module to configure and start the service.


wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
...
+                  Interactive="no"
+                  Name="pxpagent"
+                  Start="demand"
+                  Type="ownProcess"
+                  ErrorControl="normal"
+                  Vital="yes" />
+                <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\pxpagent\Parameters">
+                  <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]pxp-agent\bin" Type="expandable" />
+                  <RegistryValue Name="Application" Value="[INSTALLDIR]pxp-agent\bin\pxp-agent.exe" Type="expandable" />
+                  <RegistryValue Name="AppParameters" Value="" Type="expandable" />
```
ferventcoder: @Iristyle @parisiale @kylog Starting pxp with no parameters.

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
...
                 <File Id="NSSM" KeyPath="yes" Source="stagedir\service\nssm.exe" />
+
+                <ServiceInstall Id="PXPServiceInstaller"
+                  Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                  Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                  Description="PXP-Agent Description"
+                  DisplayName="PXP Agent"
+                  Interactive="no"
+                  Name="pxpagent"
+                  Start="demand"
```
ferventcoder: @Iristyle @parisiale @kylog Setting up as manual startup for now. We are assuming a conf file is going to get laid down in a similar manner to the way it is done for MCO. This was still a question in the ticket as of this morning, so we are taking the same approach that the configuration is going to come in from a PE module.

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
...
                 <CreateFolder />
                 <File Id="NSSM" KeyPath="yes" Source="stagedir\service\nssm.exe" />
+
+                <ServiceInstall Id="PXPServiceInstaller"
+                  Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                  Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                  Description="PXP-Agent Description"
+                  DisplayName="PXP Agent"
+                  Interactive="no"
+                  Name="pxpagent"
```
ferventcoder: We went safe on the name, removed the "-". We didn't foresee any issues with leaving it in, but a quick check through all of the installed services produced no results with a dash. So we took it out to be on the safe side.

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
               <Component Id="service_nssm" Guid="52B1CD57-95A2-4CA4-AB8E-9DDD6DE8FC66">
                 <CreateFolder />
                 <File Id="NSSM" KeyPath="yes" Source="stagedir\service\nssm.exe" />
+
```
ferventcoder: So the important thing to note here is that the next time we need an NSSM service, due to WiX fun, we have to use a different nssm executable or look at doing custom actions instead of ServiceInstall. Like we see here https://github.com/auth0/nodejs-msi/blob/master/installer/services.wxs. I'm not a fan of custom actions so I think perhaps copying that nssm.exe around for each application that will need it will be helpful so that we are using a different keypath every time. This is more FYI, we'll cross that bridge when we hit our next nssm service.

wix/puppet.wxs
joshcooper: It would be great to keep the name of the service the same everywhere, otherwise the module for managing the service needs to special case windows, e.g. `service { 'pxp-agent': ensure => running, ... }`. Based on https://msdn.microsoft.com/en-us/library/ms682450(VS.85).aspx, it appears only forward and backslashes are invalid service names.

wix/puppet.wxs
ferventcoder: @jcoconnor @kylog @parisiale @Iristyle

ferventcoder:
Minor *nit - we still need to completely fill out the Git Commit message.

wix/puppet.wxs
joshcooper: We might be able to rename the nssm binary, e.g. `pxp-agent-service.exe`?

wix/puppet.wxs
ferventcoder: Is the name of the service on other pxp-agent? I'm good to change, did not realize this.

ferventcoder: Totally forgot @joshcooper :)

wix/puppet.wxs
ferventcoder: Unless we will know server value during install, we must assume that it is going to get laid down later

wix/puppet.wxs
ferventcoder: https://github.com/puppetlabs/puppet_for_the_win/pull/126#discussion_r42139743

wix/puppet.wxs
ferventcoder: That makes sense, probably something for future reference. :)

ferventcoder: Mostly don't want to approach this at this time b/c I would foresee us copying nssm.exe to the pxp-agent bin folder with that name and then referencing it as the keypath.

ferventcoder:
Also still need to address "This ticket also includes properly configuring the pxp-module-puppet plugin with a configuration file that writes an appropriate value for puppet_bin".

ferventcoder:
I'm :+1: on these changes. @joshcooper @Iristyle do you want to give it a final once over prior to merge?

jcoconnor:
Rob,

Comments updated as requested - have also changed service name to pxp-agent.
(one minor - would be good to put in the correct PXP-Agent service description)

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
...
+                <ServiceInstall Id="PXPServiceInstaller"
+                  Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                  Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                  Description="PXP-Agent Description"
```
ferventcoder: @jcoconnor good point on this one. I can fix that up on merge. @kylog @joshcooper @parisiale @ploubser any notes on the description here? Or just something generic about "Puppet Labs Puppet Communication Protocol (PCP) service"
https://github.com/puppetlabs/puppet_for_the_win/blob/5144e4fe991f6defc99b7f31198e5d530e5590ac/wix/puppet.wxs#L228

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
...
+                  Type="ownProcess"
+                  ErrorControl="normal"
+                  Vital="yes" />
+                <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\pxp-agent\Parameters">
+                  <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]pxp-agent\bin" Type="expandable" />
+                  <RegistryValue Name="Application" Value="[INSTALLDIR]pxp-agent\bin\pxp-agent.exe" Type="expandable" />
+                  <RegistryValue Name="AppParameters" Value="" Type="expandable" />
+                  <RegistryKey Key="AppExit">
+                    <RegistryValue Value="Restart" Type="string" />
+                  </RegistryKey>
```
Iristyle: So I'm noticing a problem that we're going to have based on how the `pxp-module-puppet` module is setup.  

It performs a `require 'puppet'` at https://github.com/puppetlabs/pxp-agent/blob/master/modules/pxp-module-puppet#L5

That means it's expecting that `RUBYLIB` be populated like it has been in the MSIs `environment.bat` - https://github.com/puppetlabs/puppet_for_the_win/blob/master/conf/windows/stage/bin/environment.bat#L28

That would mean we should probably carry over other settings from `environment.bat`

However, is it really the expectation that Ruby based pxp-agent plugins have Puppet available to `require` like tha?

A couple of points:

* If the environment is expected to be available like that, then is it really necessary to shell back out to `puppet.bat` in `pxp-module-puppet`?  Seems like Puppet could be invoked more directly like https://github.com/puppetlabs/puppet/blob/master/bin/puppet.  I also don't think there would need to be a configuration value for `puppet_bin`?

* If the environment is *not* expected to be available to Ruby based `pxp-agent` modules, then does that cause issues with, for instance... resolving facts properly (a problem we've had in the past with MCO triggering `facter -p`)?  I know we haven't crossed that bridge just yet, but would be good to understand the plan.

/cc @kylog @joshcooper @richardc @ploubser @MikaelSmith

wix/puppet.wxs
ploubser: :+1: for "Puppet Labs Puppet Communication Protocol (PCP) service"

wix/puppet.wxs
richardc: So it may be that it's just very different in windows-land, but over in post puppet-agent POSIX world we don't have distinct libdirs for facter/mcollective/puppet anymore, we just install into the libdirs of the hosted ruby, so assuming you invoke `/opt/puppetlabs/puppet/bin/ruby` you get a ruby interpreter that knows where our applications are.

If that were the same for win32 then at least that section of environment.bat would be meaningless and could be excised.

So I guess my question is:  isn't it the same on win32?

ferventcoder: >However, is it really the expectation that Ruby based pxp-agent plugins have Puppet available to require like that?

I would lean towards yes.

ferventcoder: Good point, I think we'll need to bring in all of the environment variables.

MikaelSmith: `So I guess my question is: isn't it the same on win32?`

It's not. I recall some discussion of making it so, but I'm not sure what the time-frame is for that.

Iristyle: Latest thinking at time of writing is that there will be a `pxp-module-puppet.bat`, that calls `environment.bat` and which is responsible for executing `ruby pxp-module-puppet`.  Then we don't need to configure an interpreter, nor do we need to worry about adding the Ruby environment details to the service configuration.

wix/puppet.wxs
MikaelSmith: Agreed renaming makes a lot of sense (clearer for users looking at running processes).

wix/puppet.wxs
```diff
@@ -254,6 +254,28 @@
               <Component Id="service_nssm" Guid="52B1CD57-95A2-4CA4-AB8E-9DDD6DE8FC66">
                 <CreateFolder />
                 <File Id="NSSM" KeyPath="yes" Source="stagedir\service\nssm.exe" />
+
+                <ServiceInstall Id="PXPServiceInstaller"
+                  Account="[PUPPET_AGENT_ACCOUNT_DOMAIN]\[PUPPET_AGENT_ACCOUNT_USER]"
+                  Password="[PUPPET_AGENT_ACCOUNT_PASSWORD]"
+                  Description="PXP-Agent Description"
+                  DisplayName="PXP Agent"
```
Iristyle: I think that we should also prefix this like `Puppet PXP Agent` or similar... 

I know that in the past we've used `Marionette Collective`... but I think end users expect / want services from the same vendor grouped together.

jcoconnor:
I have amended the service name/description as follows:
  Description="Puppet Labs Puppet Communication Protocol (PCP) service"
  DisplayName="Puppet Labs PXP Agent"

Iristyle:
My suggestions for what I see here so far:

* I think we can shorten the `Description` / `Display Name` a bit to:
    * Puppet Communication Protocol (PCP) service
    * Puppet PXP agent
    * FWIW - I think it's a little confusing to have one thing reference PCP and the other PXP - thoughts @ploubser / @richardc ?
* Need to set the NSSM registry key that configures environment to append Ruby's bin directory to the `PATH` (typically `C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin` - but user configurable in the MSI)

ploubser:
@Iristyle I think the description should be Puppet Execution Protocol (PXP) service.

richardc:
I don't want to have more naming nightmares, so let me just state the layers:

PCP = Puppet Communications Protocol
PXP = PCP eXecution Protocol

The agent is called pxp-agent because it's an agent for handling PXP requests (which are transmitted via PCP (which is websockets and https and special sauce))

Iristyle:
@ploubser agreed - I was also thinking `Puppet Execution Protocol (PXP) Service` as well.  I was confused though by your prior :+1: to the @ferventcoder proposed name of `Puppet Labs Puppet Communication Protocol (PCP) service` on https://github.com/puppetlabs/puppet_for_the_win/pull/126#discussion_r42223739

@richardc yup, I've been explaining the same thing to everyone who keeps asking me what the acronyms mean ;0

ploubser:
@Iristyle Apologies, I misread the first time around.

wix/puppet.wxs
```diff
@@ -254,6 +254,29 @@
...
+                  Name="pxp-agent"
+                  Start="demand"
+                  Type="ownProcess"
+                  ErrorControl="normal"
+                  Vital="yes" />
+                <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\pxp-agent\Parameters">
+                  <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]pxp-agent\bin" Type="expandable" />
+                  <RegistryValue Name="Application" Value="[INSTALLDIR]pxp-agent\bin\pxp-agent.exe" Type="expandable" />
+                  <RegistryValue Name="AppParameters" Value="" Type="expandable" />
+                  <RegistryValue Name="AppEnvironmentExtra" Value="PATH=[INSTALLDIR]sys\ruby\bin" Type="multiString" Action="append" />
```
Iristyle: So if this replaces `PATH`, I'm not entirely sure this value is adequate.  I would think that we also need `C:\Windows\system32` and `c:\windows` at the very least (so that `cmd.exe` is in `PATH`).

When starting the Puppet command prompt, the following paths are in `PATH` that *could* be relevant.  I think we should shoot for the bare minimum, but I'm not entirely sure what is / is not necessary here for pxp-agent.  When the `pxp-module-puppet` plugin calls `environment.bat`, it will get most of these additions from Puppet anyway.

```
C:\Program Files\Puppet Labs\Puppet\puppet\bin
C:\Program Files\Puppet Labs\Puppet\facter\bin
C:\Program Files\Puppet Labs\Puppet\hiera\bin
C:\Program Files\Puppet Labs\Puppet\mcollective\bin
C:\Program Files\Puppet Labs\Puppet\bin
C:\Program Files\Puppet Labs\Puppet\sys\ruby\bin
C:\Program Files\Puppet Labs\Puppet\sys\tools\bin
C:\Windows\system32
C:\Windows
C:\Windows\System32\Wbem
C:\Windows\System32\WindowsPowerShell\v1.0\
```

So my vote would be to also include:

```
C:\Windows\system32
C:\Windows
C:\Windows\System32\WindowsPowerShell\v1.0\
```

ferventcoder: Surely one could append the current path to the end? `PATH=[INSTALLDIR]sys\ruby\bin;%PATH%`

I thought we removed the need for this based on conversations yesterday, perhaps I'm mistaken?

Iristyle: So even though the `AppEnvironmentExtra ` value should be a `REG_MULTI_SZ` (which normally is not expanded), there is code in NSSM to call [ExpandEnvironmentStrings](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724265\(v=vs.85\).aspx) - see commit at https://git.nssm.cc/?p=nssm.git;a=commitdiff;h=eac4d7eedfe338703f0103aac01af2ca306f615a which mentions usage of `%PATH%` specifically in the commit message.

I'm going to re-test what @jcoconnor had stated to verify actual NSSM behavior, because he seems to contradict what that commit states.

@ferventcoder - while `pxp-puppet-windows.bat` makes it unnecessary to necessarily have Ruby in the path as an interpreter, we do need to have the OpenSSL dlls (which we vendor as part of our custom built Ruby) in `PATH`, otherwise `pxp-agent.exe` crashes.

ferventcoder: Which I just saw somewhere else ha! :)

Iristyle: I verified that using `%PATH%` seems to work just fine.  Attached is a screenshot from process explorer:

screen shot 2015-10-20 at 3 37 20 pm

Iristyle:
Changes validated and merged to master in https://github.com/puppetlabs/puppet_for_the_win/commit/b515d41c89ade27d8b68d02ccbf24cf7ebed5a71

PR #128, opened 2015-10-30
(PA-64) Set pxp-agent to exit on error

jcoconnor:
Change the behaviour from auto-restart to exit so that the pxp-agent service does not restart on error exit.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -275,7 +275,10 @@
                   <!-- Skip responding to WM_QUIT and WM_CLOSE -->
                   <RegistryValue Name="AppStopMethodSkip" Value="6" Type="integer" />
                   <RegistryKey Key="AppExit">
+                    <!-- Stop the service completely on exit(2) (Invalid configuration) otherwise restart with NSSM service throttling -->
+                    <!-- nssm AppExit codes reference https://nssm.cc/usage#exit -->
                     <RegistryValue Value="Restart" Type="string" />
+                    <RegistryValue Name="2" Value="Exit" Type="string" />
                   </RegistryKey>
                 </RegistryKey>
 


ferventcoder:
Cursory :+1:

Iristyle:
So I don't think we're quite ready to merge this, pending some additional discussion on this topic.

Iristyle:
I'm running some tests now, but after some discussion I think we want to leave the default value of `Restart` and the default throttling.

With the default setup and an intentionally bad config, this is what I observed, which I think is generally acceptable:

```
- start at 9:57:34AM / enter pause state
- start at 9:57:36AM / enter pause state  + 2 sec
- start at 9:57:40AM / enter pause state  + 4 sec
- start at 9:57:48AM / enter pause state  + 8 sec
- start at 9:58:04AM / enter pause state  + 16 sec
- start at 9:58:36AM / enter pause state  + 32 sec
- start at 9:59:40AM / enter pause state  + 64 sec
- start at 10:01:48AM / enter pause state + 128 sec
- start at 10:06:04AM / enter pause state + 256 sec
- start at 10:10:20AM / enter pause state + 256 sec
- start at 10:14:36AM / enter pause state + 256 sec
- start at 10:18:52AM / enter pause state + 256 sec
- start at 10:23:08AM / enter pause state + 256 sec
- start at 10:27:24AM / enter pause state + 256 sec
- start at 10:30:29AM / STOP
```

We also have the ability to stop the service on bad config, which I think is also acceptable.  I've verified this works given `pxp-agent` will now return an exit code of `2` when the config is invalid.  In that situation, there's really nothing a user can do.  If the PE module later comes along and reconfigures pxp-agent, then it will attempt to start the service.

Using that registry setting, this is what the setup looks like:

screen shot 2015-11-04 at 2 29 58 am


Interested in additional opinions from @jpogran @jcoconnor @ferventcoder

ferventcoder:
I like option 1 or 2. However, the question of exit code 2 (bad config) begs to question how will we know that we are having issues due to a bad config and may not the wrong architecture? I'd hate to lead someone down a rabbit hole unnecessarily.

Iristyle:
I'm not sure I understand the question @ferventcoder - see https://github.com/puppetlabs/pxp-agent/pull/211/files

ferventcoder:
> leave as Restart for the default, and add Exit for exit code 2 (bad config)

> I'm not sure I understand the question

@Iristyle I think I probably misunderstood thinking this meant that we would always exit with exit code 2, where I think it is starting to solidify that we'd exit if we detected exit code 2.

ferventcoder:
@Iristyle reading this https://nssm.cc/usage#exit, I'm starting to understand what you mean. So I definitely like Option 2.

jcoconnor:
@Iristyle, @ferventcoder Updated PR with handling for Exit Code 2.

Notice though that pxp-agent still exits with 0 according to the log, so keeps restarting:

    2015-11-05 03:38:14.460418 INFO  puppetlabs.pxp_agent.util.daemonize:56 - Daemonization completed; pxp-agent PID=1728, process lock 'com_puppetlabs_pxp-agent'
    2015-11-05 03:38:14.460418 WARN  puppetlabs.pxp_agent.main:55 - pxp-agent started unconfigured. No connection can be established
    2015-11-05 05:58:34.586267 INFO  puppetlabs.pxp_agent.util.daemonize:24 - Caught signal 0 - shutting down
    2015-11-05 05:59:59.773925 INFO  puppetlabs.pxp_agent.util.daemonize:56 - Daemonization completed; pxp-agent PID=2284, process lock 'com_puppetlabs_pxp-agent'
    2015-11-05 05:59:59.789550 WARN  puppetlabs.pxp_agent.main:55 - pxp-agent started unconfigured. No connection can be established
    2015-11-05 06:32:17.887093 INFO  puppetlabs.pxp_agent.util.daemonize:24 - Caught signal 0 - shutting down
    2015-11-05 06:34:27.449860 INFO  puppetlabs.pxp_agent.util.daemonize:56 - Daemonization completed; pxp-agent PID=1236, process lock 'com_puppetlabs_pxp-agent'
    2015-11-05 06:34:27.449860 WARN  puppetlabs.pxp_agent.main:55 - pxp-agent started unconfigured.  No connection can be established

The `(Default)` string is also populated with `(value not set)` - Should I explicitly set this to `"Restart"`

jcoconnor:
Unit testing shows exit code (2) stopping the service, otherwise, service restarts.
I have explicitly set the default ExitParm to Restart.

wix/puppet.wxs
```diff
@@ -275,7 +275,10 @@
                   <!-- Skip responding to WM_QUIT and WM_CLOSE -->
                   <RegistryValue Name="AppStopMethodSkip" Value="6" Type="integer" />
                   <RegistryKey Key="AppExit">
+                    <!-- Specifying only Exit Code 2 stops the service https://nssm.cc/usage#exit -->
```
ferventcoder: The comments here are almost right. It's when the app exits with exit code 2 we want to stop the service, otherwise we allow service throttling to do its thing.

ferventcoder:
I'm still a :+1:

jcoconnor:
@ferventcoder updated comments as per your suggestion

ferventcoder:
LGTM

ferventcoder:
The ticket number in the git commit messages will need to be updated to PA-64.

jcoconnor:
done

ferventcoder:
Dropping an empty file at 
[C:\ProgramData\PuppetLabs\pxp-agent\etc\pxp-agent.conf](https://github.com/puppetlabs/pxp-agent#configuring-the-agent)

and trying to turn on the service I get:
image

jcoconnor:
@ferventcoder @Iristyle Thats expected behaviour - you can verify it form the event log.

ferventcoder:
Yep.

ferventcoder:
That's exactly what I'm looking for it to do. The next is to verify the other throttling works.

ferventcoder:
I was able to observe the behavior.

PR #130, opened 2015-11-23
(PA-127) Remove Windows 2003 deprecation warning

ferventcoder:
Now that we no longer have Windows 2003 support, we need to
back out the extra screens that would show up with the
deprecation warning.

This reverts commit 312ef72373f08dfb60c02c81638f18a9f63e47cb.

Blocked on #131


ferventcoder:
I have not done much for testing this yet, so if you pick this one up, you should validate that everything still works appropriately when you build up the bix.

glennsarti:
Tested the following;
The MSI generated ok (32/64)
ICE errors are minimal and are the same as the current agents in production (32/64)
Manually running the MSI shows the dialog doesn't appear on 2012R2 and the other dialogs aren't broken (64)
Ran the 64bit MSI on 2003 x64 and was prompted with the expected error (64)
Ran the 32bit MSI on 2003 x86 and was prompted with the expected error (32)

PR #131, opened 2015-12-21
{WIP} Merge master to stable

ferventcoder:
It was noted in #130 that the bits for not allowing the install of Windows 2003 are not in stable, I'm thinking that since we have released 4.3.0, we need to merge master back to stable. Is there any reason not to?

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -275,7 +275,10 @@
                   <!-- Skip responding to WM_QUIT and WM_CLOSE -->
                   <RegistryValue Name="AppStopMethodSkip" Value="6" Type="integer" />
                   <RegistryKey Key="AppExit">
+                    <!-- Stop the service completely on exit(2) (Invalid configuration) otherwise restart with NSSM service throttling -->
+                    <!-- nssm AppExit codes reference https://nssm.cc/usage#exit -->
                     <RegistryValue Value="Restart" Type="string" />
+                    <RegistryValue Name="2" Value="Exit" Type="string" />
                   </RegistryKey>
                 </RegistryKey>
 


ferventcoder:
@Iristyle / @melissa / @glennsarti

ferventcoder:
Wait - what? https://github.com/puppetlabs/puppet-agent/blob/stable/configs/components/windows_puppet.json#L3

Iristyle:
@ferventcoder so that's a bit confusing, but let me explain my understanding:

* puppet-agent is now releasing off of `stable` branch (i.e. no more `aardwolf`)
* puppet-agent 1.3.2 is tagged at https://github.com/puppetlabs/puppet-agent/commits/1.3.2
* If you look at the history at that point in time (based on https://github.com/puppetlabs/puppet-agent/tree/1.3.2), we can see that `puppet_for_the_win` is pointing to `refs/tags/4.3.0` - https://github.com/puppetlabs/puppet-agent/blob/1.3.2/configs/components/windows_puppet.json#L3
* Same deal for https://github.com/puppetlabs/puppet-agent/blob/1.3.0/configs/components/windows_puppet.json#L3 and https://github.com/puppetlabs/puppet-agent/blob/1.3.1/configs/components/windows_puppet.json#L3
* `1.2.7` points to `4.2.2` at https://github.com/puppetlabs/puppet-agent/blob/1.2.7/configs/components/windows_puppet.json#L3

However, we see that at present, both `master` and `stable` branches are pointing to `puppet_for_the_win` master branch.. which I don't think is what we intend.  I'm not sure why https://github.com/puppetlabs/puppet-agent/commit/0270b3a7d4712b59ee5ac2e54fe86ffd1a3ee3aa or https://github.com/puppetlabs/puppet-agent/commit/418268e30c01fbbf9928bb7d08b5c00b677c757a touched `windows_puppet.json`... but it might have been that something got flubbed when merging across master/stable in `puppet-agent`.  `puppet-agent#stable` should be tracking `puppet_for_the_win#stable` from what I last recall as our decision - any thoughts on those commits @haus ?

haus:
@Iristyle i don't remember what those were about unfortunately

Iristyle:
OK @ferventcoder I fastforwarded stable to catch up with master.  I also sent up a PR to fix the tracking - https://github.com/puppetlabs/puppet-agent/pull/463

PR #132, opened 2015-12-22
(PA-68) Filter test binaries from MSI

jpogran:
Added WIX xslt filter that ignores test binaries when building the MSI.
While this could get verbose, we can eventually split the xslt files
among the different products (i.e. one for puppet, one for facter, etc)

### DIFF
--- a/wix/filters/filters.xslt
+++ b/wix/filters/filters.xslt
@@ -14,4 +14,16 @@
   <!-- when the ruby.exe filter matches, do nothing -->
   <xsl:template match="wix:Component[wix:File[@Source='$(var.StageDir)\ruby\bin\ruby.exe']]" />
   <xsl:template match="wix:Component[wix:File[@Source='$(var.StageDir)\ruby\bin\rubyw.exe']]" />
+
+  <xsl:key name="tests-search" match="wix:Component[contains(wix:File/@Source, 'test')]" use="@Id" />
+  <xsl:template match="wix:Component[key('tests-search', @Id)]" />
+  <xsl:template match="wix:ComponentRef[key('tests-search', @Id)]" />
+
+  <xsl:key name="unittests-search" match="wix:Component[contains(wix:File/@Source, 'unittests.')]" use="@Id" />
+  <xsl:template match="wix:Component[key('unittests-search', @Id)]" />
+  <xsl:template match="wix:ComponentRef[key('unittests-search', @Id)]" />
+
+  <xsl:key name="lth_cat-search" match="wix:Component[contains(wix:File/@Source, 'lth_cat.exe')]" use="@Id" />
+  <xsl:template match="wix:Component[key('lth_cat-search', @Id)]" />
+  <xsl:template match="wix:ComponentRef[key('lth_cat-search', @Id)]" />
 </xsl:stylesheet>


wix/filters/filters.xslt
```diff
@@ -14,4 +14,16 @@
...
+  <xsl:template match="wix:ComponentRef[key('tests-search', @Id)]" />
+
+  <xsl:key name="unittests-search" match="wix:Component[contains(wix:File/@Source, 'unittests.')]" use="@Id" />
+  <xsl:template match="wix:Component[key('unittests-search', @Id)]" />
+  <xsl:template match="wix:ComponentRef[key('unittests-search', @Id)]" />
+
+  <xsl:key name="lth_cat-search" match="wix:Component[contains(wix:File/@Source, 'lth_cat.exe')]" use="@Id" />
+  <xsl:template match="wix:Component[key('lth_cat-search', @Id)]" />
+  <xsl:template match="wix:ComponentRef[key('lth_cat-search', @Id)]" />
+</xsl:stylesheet>
```
joshcooper: Ideally, we'd generate test executables in a separate directory, so that we'd just need a single rule. That would future proof changes to existing and new puppet-agent components. But that's a bit out of scope for this ticket.

ferventcoder: Is filters really the way about it? I know with Puppet et. al. we do something slightly different for removing the specs and all of the cruft.

ferventcoder: https://github.com/puppetlabs/puppet_for_the_win/blob/5e59d0e713f199253f86419ed17ca057fb9a7304/tasks/windows/windows.rake#L205-L223

ferventcoder: And I just read https://github.com/puppetlabs/puppet-agent/pull/456#issuecomment-165814508

ferventcoder: Based on what I've read, we could pull back some of the rake build items into filters, that way they can be easily ported elsewhere.

Iristyle: I was thinking the same thing about the directory approach for tests @joshcooper, which IMHO would be safer if we could get all projects agreeing to produce output there.

But for now, hopefully simple smoke / acceptance tests would pick up the accidental removal of a useful file.


@ferventcoder to your point, I believe those can be moved to the XSLT as well, yes (in another ticket)

MikaelSmith: Ideally, we'd use `make install` to gather the source for the MSI, and test files would never be installed. Which should be the future in a month or two.

Iristyle: I think to clarify on what you're stating @MikaelSmith, we would have 2 different targets at some point in the future:

* (1) zip file including application binaries (no tests)
* (2) zip file including test binaries only

(2) would only be used during a testing step of the CI pipeline, and it of course would implicitly depend on (1).  (1) is what would be consumed by packaging / etc.  The existence of (2) would then render these XSLTs unnecessary.

If that's correct, sounds good -- and something that would should probably get a ticket filed for, so we don't lose track of it.

MikaelSmith: I'm not aware of plans for a zip file including application binaries. I would expect the MSI to be created directly from an install staging area on the build machine. /cc @melissa 

Otherwise yes. But there are no tickets (yet) for any of the things you mention about testing.

melissa: @MikaelSmith there will be no zip files for application binaries. We will use logic in vanagon to stage all the files produced from the `make install`, and build the msi with those files. We will not create a zip archive for public consumption in any part of the process beyond the nuget/msi packages.

@Iristyle where did the zip file including application binaries (no tests) come from?

Iristyle: @melissa we need to be able to verify packaging changes without running the full Vanagon process end-to-end.  This is why there are currently intermediate assets like `pxp-agent-1.3.2.149.g9dab60a-x64.zip` and `facter-1.3.2.149.g9dab60a-x64.zip` produced by the builds in a SHA based location like http://builds.delivery.puppetlabs.net/puppet-agent/9dab60a40f73296a3499bcffc19e8e32b2ef9b2d/artifacts/windows/

If we can't do ad-hoc local builds of the packaging without rebuilding the whole machine, then that's a significant step backwards IMHO.

MikaelSmith: There is a workflow that will allow iterating on MSI packaging with Vanagon, but it requires one full Vanagon build to setup. If you preserve the build environment you can then go in, make changes to the MSI config, and use `make` to repeat packaging.

Iristyle: @MikaelSmith unfortunately I think requiring a full Vanagon run first to be able to iterate on packaging changes is not going to work.

Unless Vanagon can use Vagrant / VMWare / VirtualBox instead of vmpooler, that requires a local or VPN connection (which can be bad based on your physical location or if you're offline) ... and it's prohibitively slow to initially setup.  It also doesn't scale well to support multiple concurrent users due to the potential for pooler exhaustion.

melissa: The pooler is one of many options :) All of the above are options for use with vanagon. You can even specify a machine to use via the hostname and it will use that. I think a hardware engine was just merged in as well.

Also, I wouldn't be too worried about pooler exhaustion. It does take a little longer to spin up windows machines there, but we can always increase the pool size if it becomes an issue.

Iristyle:
This was reverted since it excluded 300+ files that shouldn't have been excluded.

PR #133, opened 2015-12-30
(PA-68) Filter test assemblies for Puppet MSI

jpogran:
Use a WiX xslt filter to ignore the following files:

lth_cat.exe
libfacter_test.exe
libtest.so
libtest1.so
cpp-pcp-client-unittests.exe
pxp-agent-unittests.exe

The first attempt at this removed them from the staging directories
inside ther rake task for building the agent itself. This was decided to
be too heavy handed, as end users may want the test assemblies in
downstream CI pipelines.

The second attempt used WiX filters to ignore the specified files when generating
WiX fragment files from the staging directory. This was a better
approach in that it left the files for anyone who needed them, and
removed them on the packaging step where they were not wanted. However
the filters turned out to be too greedy, and removed hundreds of files
from the MSI.

The third attempt uses a stricter filter that only removes files
matching the names provided. There is too much overlap with words like
'test' to rely on simple matches. The better way to have approached this
is to have an xslt for each product (one for puppet, one for facter,
etc...) however this would require changing the rake tasks significantly
and would probably be superseded by vanagaon.

### DIFF
--- a/wix/filters/filters.xslt
+++ b/wix/filters/filters.xslt
@@ -14,4 +14,12 @@
   <!-- when the ruby.exe filter matches, do nothing -->
   <xsl:template match="wix:Component[wix:File[@Source='$(var.StageDir)\ruby\bin\ruby.exe']]" />
   <xsl:template match="wix:Component[wix:File[@Source='$(var.StageDir)\ruby\bin\rubyw.exe']]" />
+
+  <!-- Ignore test assemblies  -->
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'lth_cat.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libfacter_test.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libtest.so')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libtest1.so')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'cpp-pcp-client-unittests.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'pxp-agent-unittests.exe')]" />
 </xsl:stylesheet>


wix/filters/filters.xslt
```diff
@@ -14,4 +14,12 @@
...
-</xsl:stylesheet>
\ No newline at end of file
+
+  <!-- Ignore test assemblies  -->
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'lth_cat.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libfacter_test.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libtest.so')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'libtest1.so')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'cpp-pcp-client-unittests.exe')]" />
+  <xsl:template match="wix:Component[contains(wix:File/@Source, 'pxp-agent-unittests.exe')]" />
```
ferventcoder: Are these all in the same folder? I'm guessing no.

ferventcoder:
Minor *nit - your commit message should probably state which attempt this is. And for the second attempt, you may want to include the commit sha and the reverted commit sha.

wix/filters/filters.xslt
jpogran: The ticket lists which ones are in which directory, two directories in total

jpogran:
@Iristyle @ferventcoder @glennsarti The alternative approach is to use full paths like line 15 and 16. I originally did not use this approach because we were trying to use matches and not direct paths to reduce the amount of lines we needed to write. But since the simple matches we are using match more than what we want, we are back to one line per exe which might as well be a full path. Also Glenn thought a prefixed exe (foo.something.exe where we specify something.exe) would produce false matches. This is supported by lth_cat.exe matching for both pxp-agent and facter, which is desirable now but could be undesirable for something else that has a similar name to what we have here. What do you think?

glennsarti:
I'm leaning towards one line per file to exclude.  It does mean a little duplication but it's very explicit and easy to read.  Also it may be easier later to refactor into excludes per component.

ferventcoder:
We use glob filtering in the rake task, so I tend to lean towards that. Since we do extensive testing of the package, if the name matches we may want it removed. But I'm 6 to 1, half dozen to the other.

ferventcoder:
Verifying this now.

ferventcoder:
image


Looks good.

Iristyle:
This is a redo for #132

Iristyle:
#134 additionally removes `libcpp-pcp-client-unittests.dll.a`

Iristyle:
Generally speaking, I'd rather see us exclude a well-defined folder from each native project, so that we can just set it and forget it here.  Otherwise, every time new test artifacts are produced, we end up requiring changes here.

Thoughts on unifying and making test binary output consistent across native projects @kylog / @MikaelSmith ?

MikaelSmith:
This will be addressed in about a month by Vanagon work. The layout of inputs to puppet_for_the_win are determined by the powershell scripts, and modifying those was the prior attempt to this PR. Using `make install` there to isolate required files would've been the cleaner approach you're asking for.

Within component build directories, it's convenient for all binary files to be in the same directory to make running executables simpler.

PR #135, opened 2016-02-08
(RE-5747) Update signtool to use SHA-256

geoffnichols:
In support of the upcoming transition to signing Windows
MSIs with a SHA-256 code signing cert, we update the
signtool flag to specify use of the SHA-256 file digest
algorithm.

(Prior to this commit, signtool uses the default SHA-1
algorithm.)


McdonaldSeanp:
:+1:

joshcooper:
According to this [post](http://stackoverflow.com/questions/18542160/signtool-failing-to-dual-sign-sha2-and-sha1-with-timestamps), 2008 (non-R2) and Vista do not accept sha256 certs. It indicates you can create two signatures, but requires having two signing certs. Not sure what we want to do? @Iristyle

geoffnichols:
@joshcooper @e Thanks, I definitely appreciate your guidance here. One note - I _think_ there may be a distinction between requirements for .dll vs .msi signing, but I find the documentation a bit confusing.

geoffnichols:
Hmm, after reading [this](http://social.technet.microsoft.com/wiki/contents/articles/32288.windows-enforcement-of-authenticode-code-signing-and-timestamping.aspx#How_to_handle_MSI_files) I think this PR may be unnecessary?

PR #136, opened 2016-02-19
(PUP-5933) Don't set pluginsync in puppet.conf

whopper:
As of Puppet 4.4.0, setting the pluginsync option by CLI
or config is deprecated and will log a deprecation warning,
as that behavior will eventually be managed based on whether or not
we are using a cached catalog.

While the setting is deprecated, it still defaults to true,
so removing this setting in puppet.conf will not cause a
change in behavior.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -325,11 +325,6 @@
                   Section="main"
                   Key="server" Value="[PUPPET_MASTER_SERVER]"
                   Directory="PuppetConfDir" />
-                <IniFile Id="PuppetConfPluginSync" Name="puppet.conf"
-                  Action="createLine"
-                  Section="main"
-                  Key="pluginsync" Value="true"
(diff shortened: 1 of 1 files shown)


whopper:
I looked at doing a test build with this change, but realized I don't actually know how to build MSIs :(

whopper:
Going to make a new ticket for this.

ferventcoder:
LGTM!

ferventcoder:
@whopper note that this will not remove it from existing installations of Puppet. The value in the conf file will still be there. So you may need to do something additional to remove that (which maybe is what the deprecation warning is for).

whopper:
@ferventcoder yeah, good point. Hopefully it won't be too annoying for folks

Iristyle:
This ticket has been moved to [PA-226](https://tickets.puppetlabs.com/browse/PA-226)

PR #137, opened 2016-02-23
(PA-231) Keep Existing Puppet.Conf Settings

ferventcoder:
In `#20281` (167999a13afd622b0bc823650efbe303106b5ad0), we made a
change to allow overriding the current puppet.conf settings. This
came out in Puppet 3.4.0. This allowed being able to pass updated
settings to the installer and having them persist to the conf
file. Unfortunately, that also meant that any changes to the
puppet.conf file outside of install/upgrade time would be ignored
and overwritten by the remembered property values, which is
undesirable.

The conf file should also be looked at as an additional source of
remembered properties to allow keeping those settings when a user
is not overriding those settings during the install/upgrade. To do
this we need to copy the conf file to the Windows directory, read
the property values, and then remove the file from the Windows
directory when we are done.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -614,6 +614,59 @@
         Name="RememberedPuppetAgentStartupMode" Type="raw" Win64="no" />
     </Property>
 
+    <!-- Properties for existing Ini Values - expanding the remembered properties pattern just a bit
+    -->
+
+    <CustomAction Id="SetIniPropertyValues"
+                  Script="vbscript"
+                  Execute="firstSequence"
+                  Return="ignore">
+      <![CDATA[
+On Error Resume Next
+
+Function SetPropertyFromIni (keyName, propertyName, textToSearch)
+  Set iniValueRegex = New RegExp
+  iniValueRegex.IgnoreCase = true
+  iniValueRegex.Multiline = true
+  iniValueRegex.Global = false
+  iniValueRegex.Pattern = "^" & keyName & "=(.*[^\s\r\n]*)$"
+
+  Set matches = iniValueRegex.Execute(textToSearch)
+
+  dim iniValue
+  iniValue = ""
+
+  If (matches.Count <> 0) Then
+    If (matches.Item(0).SubMatches.Count <> 0) Then
+      iniValue = matches.Item(0).SubMatches(0)
+      iniValue = Replace(Replace(Replace(iniValue, Chr(13), ""), vbNewLine, ""), vbLf, "")
+    End If
+  End If
+
+  If (iniValue <> "") Then
+    Session.Property(propertyName) = iniValue
+  End If
+End Function
+
+Dim fso, wshShell, iniPath
+Set wshShell = CreateObject("WScript.Shell")
+iniPath = wshShell.ExpandEnvironmentStrings("%ALLUSERSPROFILE%\PuppetLabs\puppet\etc\puppet.conf")
+
+Set fso = CreateObject("Scripting.FileSystemObject")
+If (fso.FileExists(iniPath)) Then
+  Set iniFile = fso.OpenTextFile(iniPath, 1, false)
+  iniFileText = iniFile.ReadAll()
+  iniFile.Close
+  Set iniFile = Nothing
+
+  SetPropertyFromIni "server", "INI_PUPPET_MASTER_SERVER", iniFileText
+  SetPropertyFromIni "environment", "INI_PUPPET_AGENT_ENVIRONMENT", iniFileText
+  SetPropertyFromIni "certname", "INI_PUPPET_AGENT_CERTNAME", iniFileText
+  SetPropertyFromIni "ca_server", "INI_PUPPET_CA_SERVER", iniFileText
+End If
+      ]]>
+    </CustomAction>
+
     <!-- Custom Actions to handle command line property values that override
          remembered property values -->
     <!-- INSTALLDIR -->
@@ -626,6 +679,10 @@
       Value="[CMDLINE_INSTALLDIR]"
       Execute="firstSequence" />
     <!-- PUPPET_MASTER_SERVER -->
+    <CustomAction Id="SetFromIniPuppetMasterServer"
+      Property="PUPPET_MASTER_SERVER"
+      Value="[INI_PUPPET_MASTER_SERVER]"
+      Execute="firstSequence" />
     <CustomAction Id="SaveCmdLinePuppetMasterServer"
       Property="CMDLINE_PUPPET_MASTER_SERVER"
       Value="[PUPPET_MASTER_SERVER]"
@@ -639,6 +696,10 @@
       Value="puppet"
       Execute="firstSequence" />
     <!-- PUPPET_AGENT_ENVIRONMENT -->
+    <CustomAction Id="SetFromIniPuppetAgentEnvironment"
+      Property="PUPPET_AGENT_ENVIRONMENT"
+      Value="[INI_PUPPET_AGENT_ENVIRONMENT]"
+      Execute="firstSequence" />
     <CustomAction Id="SaveCmdLinePuppetAgentEnvironment"
       Property="CMDLINE_PUPPET_AGENT_ENVIRONMENT"
       Value="[PUPPET_AGENT_ENVIRONMENT]"
@@ -652,6 +713,10 @@
       Value="production"
       Execute="firstSequence" />
     <!-- PUPPET_AGENT_CERTNAME -->
+    <CustomAction Id="SetFromIniPuppetAgentCertname"
+      Property="PUPPET_AGENT_CERTNAME"
+      Value="[INI_PUPPET_AGENT_CERTNAME]"
+      Execute="firstSequence" />
     <CustomAction Id="SaveCmdLinePuppetAgentCertname"
       Property="CMDLINE_PUPPET_AGENT_CERTNAME"
       Value="[PUPPET_AGENT_CERTNAME]"
@@ -661,6 +726,10 @@
       Value="[CMDLINE_PUPPET_AGENT_CERTNAME]"
       Execute="firstSequence" />
     <!-- PUPPET_CA_SERVER -->
+    <CustomAction Id="SetFromIniPuppetCaServer"
+      Property="PUPPET_CA_SERVER"
+      Value="[INI_PUPPET_CA_SERVER]"
+      Execute="firstSequence" />
     <CustomAction Id="SaveCmdLinePuppetCaServer"
       Property="CMDLINE_PUPPET_CA_SERVER"
       Value="[PUPPET_CA_SERVER]"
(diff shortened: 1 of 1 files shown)


ferventcoder:
So this mostly works. It will not copy the ini file to the Windows directory properly though.

It does have the additional bonus of updating the remembered properties in the registry.

wix/puppet.wxs
```diff
@@ -614,6 +614,51 @@
...
+    -->
+
+    <CustomAction Id="CopyPuppetConfFileWinDir_SetProp"
+              Property="CopyPuppetConfFileWinDir"
+              Value="&quot;[%WINDIR]\System32\WindowsPowerShell\v1.0\powershell.exe&quot; -NoLogo -NonInteractive -InputFormat None -NoProfile -ExecutionPolicy Bypass -Command &quot;try [\{] Copy-Item &apos;[%ALLUSERSPROFILE]\PuppetLabs\puppet\etc\puppet.conf&apos; &apos;[%WINDIR]\puppet.conf&apos; -Force  [\}] finally [\{] [\}]&quot;"
+              Execute="immediate" />
+    <CustomAction Id="CopyPuppetConfFileWinDir"
+              BinaryKey="WixCA"
+              DllEntry="CAQuietExec"
+              Execute="immediate"
```
ferventcoder: As "immediate" I get `CAQuietExec: error 0x80070057: failed to get command line` and it won't copy. As "deferred", I get an issue with the location of the sequence (currently after AppSearch).

ferventcoder: immediate:


image

ferventcoder: deferred:

~~~sh
C:\codelocal\puppetwinbuilder\src\puppet_for_the_win\wix\puppet.wxs(844) : error LGHT0204 : ICE77: CopyPuppetConfFileWinDir is a in-script custom action.  It must be sequenced in between the InstallInitialize action and the InstallFinalize action in the InstallExecuteSequence table
rake aborted!
Command failed with status (204): [light -ext WiXUtilExtension -ext WixUIExte...]
~~~

wix/puppet.wxs
```diff
@@ -614,6 +614,51 @@
...
+    </Property>
+
+    <!-- Custom Actions for putting the ini file in the Windows Directory - required to be there by IniLocator - see https://msdn.microsoft.com/en-us/library/aa369283.aspx -->
+    <!--
+
+       "c:\windows\System32\WindowsPowerShell\v1.0\powershell.exe" -NoLogo -NonInteractive -InputFormat None -NoProfile -ExecutionPolicy Bypass -Command "$returnValue = 'puppet'; try { Get-Content '%AllUsersProfile%\PuppetLabs\puppet\etc\puppet.conf' | ?{ $_ -match '^server=(?<value>.*[^\s\t\r\n]*)' } | %{ $matches['value'] } | %{$returnValue= $_} } finally { } return $returnValue;"
+       server=puppet
+       &quot;[%WINDIR]\System32\WindowsPowerShell\v1.0\powershell.exe&quot; -NoLogo -NonInteractive -InputFormat None -NoProfile -ExecutionPolicy Bypass -Command &quot;$returnValue=&apos;puppet&apos;; try [\{] Get-Content &apos;[%ALLUSERSPROFILE]\PuppetLabs\puppet\etc\puppet.conf&apos; | [\?][\{] $_ -match &apos;[\^]server=([\?]&lt;value&gt;.*[\[][\^]\s\t\r\n[\]]*)&apos; [\}] | %[\{] $matches[\[]&apos;value&apos;[\]] [\}] | %[\{]$returnValue= $_[\}] [\}] finally [\{] [\}] return $returnValue;&quot;
```
ferventcoder: This was for an original try of setting a property to the output of an exec from the file without having to move it, but I couldn't determine how to set a property value from the output of an exec.

wix/puppet.wxs
```diff
@@ -614,6 +614,51 @@
...
+
+    <CustomAction Id="DeletePuppetConfFileWinDir_SetProp"
+             Property="DeletePuppetConfFileWinDir"
+             Value="&quot;[%WINDIR]\System32\WindowsPowerShell\v1.0\powershell.exe&quot; -NoLogo -NonInteractive -InputFormat None -NoProfile -ExecutionPolicy Bypass -Command &quot;try [\{] Remove-Item &apos;[%WINDIR]\puppet.conf&apos; -Force  [\}] finally [\{] [\}]&quot;"
+             Execute="immediate" />
+    <CustomAction Id="DeletePuppetConfFileWinDir"
+              BinaryKey="WixCA"
+              DllEntry="CAQuietExec"
+              Execute="deferred"
+              Return="ignore" />
```
ferventcoder: We may be able to move this to a component instead of a custom action to ensure the file is removed.

wix/puppet.wxs
ferventcoder: The problem is that it needs to be earlier in the process due to gathering the value from the INI prior to setting the value that has been passed into the MSI - https://msdn.microsoft.com/en-us/library/windows/desktop/aa372038(v=vs.85).aspx

ferventcoder:
@puppetlabs/windows poke holes into this. There is one area that is not solid and that is getting puppet.conf into the Windows directory early in the process to gather the values from it.

ferventcoder:
[puppet-agent-x64.msi.zip](https://github.com/puppetlabs/puppet_for_the_win/files/143399/puppet-agent-x64.msi.zip)

ferventcoder:
We are down to a couple of options. 

* We can do a VBS custom action and embed it directly in. This is the only way you can set an MSI property.
* We can try going back down my first route, except this time set the values in the remembered property values in the registry.

ferventcoder:
Moved from PUP-5769 to PA-231.

wix/puppet.wxs
Iristyle: /cc @jcoconnor for any Wix sequence insight here...

ferventcoder: After further research, it is that we don't yet have permission to copy to the Windows folder, which we won't get until it moves PAST the UI to perform the actual install. That is too late in the process to be useful.

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
         Name="RememberedPuppetAgentStartupMode" Type="raw" Win64="no" />
     </Property>
 
+    <!-- Properties for existing Ini Values - expanding the remembered properties pattern just a bit
+    -->
+
+    <CustomAction Id="SetIniPropertyValues"
+                  Script="vbscript"
+                  Execute="firstSequence">
```
ferventcoder: Going to add return ignore here so if it fails, that's okay.

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
...
+
+  'wscript.Echo iniFileText
+
+  SetPropertyFromIni "server", "INI_PUPPET_MASTER_SERVER", iniFileText
+  SetPropertyFromIni "environment", "INI_PUPPET_AGENT_ENVIRONMENT", iniFileText
+  SetPropertyFromIni "certname", "INI_PUPPET_AGENT_CERTNAME", iniFileText
+  SetPropertyFromIni "ca_server", "INI_PUPPET_CA_SERVER", iniFileText
+End If
+      ]]>
+    </CustomAction>
```
ferventcoder: I have stooped low on this one. :cry:

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
...
+  dim iniValue
+
+  If matches.Count <> 0 Then
+    If matches.Item(0).SubMatches.Count <> 0 Then
+      iniValue = matches.Item(0).SubMatches(0)
+      iniValue = Replace(Replace(Replace(iniValue, Chr(13), ""), vbNewLine, ""), vbLf, "")
+    End If
+  End If
+
+  If iniValue <> Empty Then
```
glennsarti: Shouldn't you use  If Not IsEmpty(iniValue) then

ferventcoder: Probably. My VB is a bit rusty. I knew I didn't want Null.

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
...
+    'WScript.Echo "Value for " & keyName & " of '" & iniValue & "' will be set to " & propertyName
+    Session.Property(propertyName) = iniValue
+  End If
+End Function
+
+Dim fso, wshShell, iniPath
+Set wshShell = WScript.CreateObject("WScript.Shell")
```
glennsarti: Not sure if you can use `WScript.`  in a hosted MSI session.

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
...
+Function SetPropertyFromIni (keyName, propertyName, textToSearch)
+  Set iniValueRegex = New RegExp
+  iniValueRegex.IgnoreCase = true
+  iniValueRegex.Multiline = true
+  iniValueRegex.Global = false
+  iniValueRegex.Pattern = "^" & keyName & "=(.*[^\s\r\n]*)$"
+
+  Set matches = iniValueRegex.Execute(textToSearch)
+
+  dim iniValue
```
glennsarti: You haven't set Option Explicit so Dim-ing isn't really all that necessary

wix/puppet.wxs
```diff
@@ -614,6 +614,58 @@
...
+    End If
+  End If
+
+  If iniValue <> Empty Then
+    'WScript.Echo "Value for " & keyName & " of '" & iniValue & "' will be set to " & propertyName
+    Session.Property(propertyName) = iniValue
+  End If
+End Function
+
+Dim fso, wshShell, iniPath
```
glennsarti: Not sure if you want to trap or ignore errors from here on (On Error Resume ....)

wix/puppet.wxs
glennsarti: Did you still need the commented code in there?

wix/puppet.wxs
glennsarti: Might be easier to just assign a value to it (e.g. "") prior to the if and then check if it's blank.  Empty != Null != Nothing

wix/puppet.wxs
ferventcoder: Probably not. Will remove that.

wix/puppet.wxs
ferventcoder: Yep.

wix/puppet.wxs
ferventcoder: Or Option Strict... man you are making it all come back now... haha

wix/puppet.wxs
ferventcoder: Ignore them really.

wix/puppet.wxs
ferventcoder: Removed.

ferventcoder:
This is good to go.

wix/puppet.wxs
```diff
@@ -614,6 +614,59 @@
...
+  iniValue = ""
+
+  If matches.Count <> 0 Then
+    If matches.Item(0).SubMatches.Count <> 0 Then
+      iniValue = matches.Item(0).SubMatches(0)
+      iniValue = Replace(Replace(Replace(iniValue, Chr(13), ""), vbNewLine, ""), vbLf, "")
+    End If
+  End If
+
+  If Not IsEmpty(iniValue) Then
```
glennsarti: I don't think you want to use IsEmpty here.  Blank/empty strings are not Empty.  Wouldn't it be better to use `If (iniValue <> "")` then ....

ferventcoder: Doh. You asked me to change that. Lulz

ferventcoder: So if I remove line 637, it should work again?

wix/puppet.wxs
```diff
@@ -614,6 +614,59 @@
...
+  End If
+
+  If (iniValue <> "") Then
+    Session.Property(propertyName) = iniValue
+  End If
+End Function
+
+Dim fso, wshShell, iniPath
+Set wshShell = CreateObject("WScript.Shell")
+iniPath = wshShell.ExpandEnvironmentStrings("%ALLUSERSPROFILE%\PuppetLabs\puppet\etc\puppet.conf")
```
Iristyle: Do we know the difference between this, and something like

```vbscript
wshShell.Environment("Process").Item("ALLUSERSPROFILE") & "\PuppetLabs\puppet\etc\puppet.conf"
```

Also, I don't suppose there's a way to grab the variable for location of `puppet.conf`?

glennsarti: AFAIK the major difference would be if ALLUSERSPROFILE didn't exist.  In your example an error would probably occur (non existent key) whereas the code in the commit would result in iniPath being set to `%ALLUSERSPROFILE%\....`.  The code in the commit is the canonical way of determining paths via environment variables.

There may well be a way to get a non-static way for getting the puppet.conf but I'm not sure it really matters.  Is there even a way to change the path that puppet itself uses for puppet.conf or is it effectively hard set to puppet/etc/conf/puppet.conf?

ferventcoder: @Iristyle I looked at both ways and they were similar. I didn't find much documentation related to the recommended way, so I stuck with the ExpandEnvironmentStrings.

ferventcoder: As far as grabbing the output location for it, I think it might be possible by scanning the MSI database but I'm not sure it is worth the effort.

PR #138, opened 2016-03-07
(PA-225) Remove pluginsync setting from conf

ferventcoder:
During upgrade scenarios, if the pluginsync value is found,
it should be removed from the puppet.conf file. This is follow up
for a7f45d46e018038b26dfd653465b6f581d27e1fe, which deprecated
setting the pluginsync value.


ferventcoder:
This removes the value from the file if it is there.

ferventcoder:
This also ignores the value if it is missing. Exactly what we are looking for and for once a straightforward change.

PR #139, opened 2016-03-14
(PCP-276) Don't use ProcessTerminate in pxp-agent nssm

parisiale:
Using only CTRL-C, not TerminateProcess, to avoid killing the entire
precess tree. Setting the AppKillProcessTree flag to 0.

With this commit, we configure the pxp-agent service for nssm with the
above settings, to prevent killing the entire pxp-agent process tree
once a "service stop" is triggered. Note that the AppKillProcessTree
alone is not sufficient to avoid terminating the entire process tree.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -266,14 +266,16 @@
                   Type="ownProcess"
                   ErrorControl="normal"
                   Vital="yes" />
-                <!-- Various registry keys documented at https://nssm.cc/usage -->
+                <!-- Various registry keys documented at https://nssm.cc/usage and https://github.com/kirillkovalenko/nssm -->
                 <RegistryKey Root="HKLM" Key="SYSTEM\CurrentControlSet\Services\pxp-agent\Parameters">
                   <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]pxp-agent\bin" Type="expandable" />
                   <RegistryValue Name="Application" Value="[INSTALLDIR]pxp-agent\bin\pxp-agent.exe" Type="expandable" />
                   <RegistryValue Name="AppParameters" Value="" Type="expandable" />
                   <RegistryValue Name="AppEnvironmentExtra" Value="PATH=[INSTALLDIR]sys\ruby\bin;%PATH%" Type="multiString" Action="append" />
-                  <!-- Skip responding to WM_QUIT and WM_CLOSE -->
-                  <RegistryValue Name="AppStopMethodSkip" Value="6" Type="integer" />
+                  <!-- Skip responding to WM_QUIT and WM_CLOSE; don't use TerminateProcess to prevent killing the entire process tree (PCP-276) -->
+                  <RegistryValue Name="AppStopMethodSkip" Value="14" Type="integer" />
+                  <!-- Set the flag for not killing the entire process tree (PCP-276) -->
+                  <RegistryValue Name="AppKillProcessTree" Value="0" Type="integer" />
                   <RegistryKey Key="AppExit">
                     <!-- Stop the service completely on exit(2) (Invalid configuration) otherwise restart with NSSM service throttling -->
                     <!-- nssm AppExit codes reference https://nssm.cc/usage#exit -->


wix/puppet.wxs
```diff
@@ -266,14 +266,16 @@
...
                   <RegistryValue Name="AppDirectory" Value="[INSTALLDIR]pxp-agent\bin" Type="expandable" />
                   <RegistryValue Name="Application" Value="[INSTALLDIR]pxp-agent\bin\pxp-agent.exe" Type="expandable" />
                   <RegistryValue Name="AppParameters" Value="" Type="expandable" />
                   <RegistryValue Name="AppEnvironmentExtra" Value="PATH=[INSTALLDIR]sys\ruby\bin;%PATH%" Type="multiString" Action="append" />
-                  <!-- Skip responding to WM_QUIT and WM_CLOSE -->
-                  <RegistryValue Name="AppStopMethodSkip" Value="6" Type="integer" />
+                  <!-- Skip responding to WM_QUIT and WM_CLOSE; don't use TerminateProcess to prevent killing the entire process tree (PCP-276) -->
+                  <RegistryValue Name="AppStopMethodSkip" Value="14" Type="integer" />
+                  <!-- Set the flag for not killing the entire process tree (PCP-276) -->
+                  <RegistryValue Name="AppKillProcessTree" Value="0" Type="integer" />
```
ferventcoder: :+1:

ferventcoder:
@puppetalbs/windows @jcoconnor FYI.

ferventcoder:
@puppetlabs/windows FYI

jcoconnor:
:+1:
Need to track this for the Windows/Vanagon work

joshcooper:
If you execute `net stop pxp-agent`, I would expect the entire pxp-agent process tree to exit, but with this PR, only the pxp-agent service would exit?

If we don't stop the child processes, then it can cause problems during upgrades. For example, if you install a new MSI, it will stop the pxp-agent service, remove the old files, install the new files, and start the service back up. But if we don't kill the process tree, then the upgrade will trigger "reboot is required" for any executable file (`exe`, `dll`, etc)  that is currently running.

Puppet has this problem currently, and we just haven't updated its service code to be smarter. For example, we don't ensure ruby exits, which can cause problems when we try to install a new version during an upgrade, e.g. ruby 2.1.7 to 2.1.8.

Is the original issue solved completely by PCP-276? In other words, if pxp-agent executes child module process in a different process group, not part of the pxp-agent parent job, then why do we need this PR?

parisiale:
No, the original issue is not solved by just using `CREATE_NO_WINDOW | CREATE_NEW_PROCESS_GROUP` when we call `CreateProcess`.

joshcooper:
Ok I think I understand now. With this change, `net stop pxp-agent` will stop the service and all child processes associated with blocking pxp-agent requests, but not processes associated with non-blocking requests.

For non-blocking requests, pxp-agent executes the child process in a [new process group](https://github.com/parisiale/leatherman/blob/9824b5ac481b2492f5f62ee873a5e0c314e140c1/execution/src/windows/execution.cc#L543-L545). MSDN says the following about the [`CREATE_NEW_PROCESS_GROUP`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684863(v=vs.85).aspx) flag:

> If this flag is specified, CTRL+C signals will be disabled for all processes within the new process group.

Previously, PFTW used [`AppStopMethodSkip 6`](https://github.com/kirillkovalenko/nssm/blob/ffbcd9c99074239f1edd1a7d9a9a6c816ff893d8/README.txt#L239-L242), so it would try to stop child processes using either [`CTRL_C_EVENT`](https://github.com/kirillkovalenko/nssm/blob/ffbcd9c99074239f1edd1a7d9a9a6c816ff893d8/process.cpp#L251) or `TerminateProcess`. This PR eliminates the use of `TerminateProcess`, and since child processes in the new process group ignore `CTRL_C_EVENT`, they continue running after the pxp-agent service exits.

I see two possible downsides to this change:

1. If an executable from a blocking request doesn't respond to the `CTRL_C_EVENT` in a timely fashion (default 1.5 secs), then it won't be terminated. Will that cause the service to fail to stop? Or will the process run in the background indefinitely? See also https://github.com/kirillkovalenko/nssm/blob/master/README.txt#L248-L250.
1. The above situation can easily occur if the executable is GUI-based, e.g. `rubyw.exe`, since they don't allocate a console by default, and so don't receive the Ctrl-C. Prior to this PR, GUI applications would have received the `WM_CLOSE`/`WM_QUIT` and could shutdown that way. This scenario has me a little concerned because it means if pxp-agent executes any GUI applications, perhaps indirectly via external facts, then that child process(es) won't be terminated when pxp-agent is shutdown. Maybe it's always short-lived, so we don't care?

Ideally, nssm would only called TerminateProcess on children in the service's process group. That way we'd always terminate processes associated with blocking requests, while ignoring processes from non-blocking requests. But that's a bit outside the scope of this ticket I think.

Given all the options, I'm :+1: on this.

MikaelSmith:
If those other scenarios comes up, it sounds like our only recourse is to write an actual service and stop using NSSM.

ferventcoder:
> If those other scenarios comes up, it sounds like our only recourse is to write an actual service and stop using NSSM.

:/

ferventcoder:
We could also fix NSSM.

Iristyle:
Yeah, I think our service usage here is specific enough at this point to warrant a real Windows service that allows us to control its behavior explicitly.  The idea of a generic supervisor is nice, but...

Iristyle:
In thinking about the concept here presented a bit more, I'm not sure that servicing non-blocking requests with a detached child process is a good idea (not just on Windows, but all platforms - though I think it's potentially more problematic on Windows).

I'm concerned about zombies, general process management, and upgrades / locked files.  I'm not clear that the argument for detached children is compelling enough to warrant the additional complexity / runtime issues that this might cause.

So my position at this point is that this PR should be closed, and that current behavior should be preserved.

Iristyle:
@parisiale / @jcoconnor / @richardc / @james-stocks I think we're going to close this and maintain existing behavior - but if anyone has any objections, please feel free to comment on the existing e-mail discussion.


Thanks!

parisiale:
@Iristyle PCP-276 is now closed as "won't do" :+1:

wix/puppet.wxs
Iristyle: https://github.com/kirillkovalenko/nssm/blob/95640f304295257c6d1bb18c806f9d1feb714d61/README.txt#L240-L248

PR #142, opened 2016-05-05
(PA-286) remove default environment setting in puppet.conf

McdonaldSeanp:
In the current windows puppet setup, when installing the agent, the MSI will
update the puppet.conf file with environment=production by default. this will
break the workflow described in
https://docs.puppet.com/pe/latest/console_classes_groups_environment_override.html#workflow
so we need to update the MSI so it sets no default value.

### DIFF
--- a/wix/puppet.wxs
+++ b/wix/puppet.wxs
@@ -708,10 +708,6 @@ End If
       Property="PUPPET_AGENT_ENVIRONMENT"
       Value="[CMDLINE_PUPPET_AGENT_ENVIRONMENT]"
       Execute="firstSequence" />
-    <CustomAction Id="SetDefaultPuppetAgentEnvironment"
-      Property="PUPPET_AGENT_ENVIRONMENT"
-      Value="production"
-      Execute="firstSequence" />
     <!-- PUPPET_AGENT_CERTNAME -->
     <CustomAction Id="SetFromIniPuppetAgentCertname"
       Property="PUPPET_AGENT_CERTNAME"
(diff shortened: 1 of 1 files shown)


ferventcoder:
@McdonaldSeanp do you need it actually removed from the conf?

wix/puppet.wxs
```diff
@@ -874,9 +867,6 @@ End If
       <Custom Action='SetFromCmdLinePuppetAgentEnvironment' After='FileCost'>
         CMDLINE_PUPPET_AGENT_ENVIRONMENT
       </Custom>
-      <Custom Action='SetDefaultPuppetAgentEnvironment' Before='CostFinalize'>
-        PUPPET_AGENT_ENVIRONMENT=""
-      </Custom>
```
ferventcoder: This probably has the side effect of not setting the value in the puppet.conf since it will be blank by default. Are we allowing folks to specify on the command line still?

ferventcoder: So now that I understand this better, this is strictly about removing production by default?

ferventcoder:
https://github.com/puppetlabs/puppet_for_the_win/blob/e724e2d4d00f56fbf40328726cb663d9f1eb413f/wix/puppet.wxs#L341-L345I think we are good. If the value is empty, it will be production by default (but won't set the conf setting).

ferventcoder:
@puppetlabs/windows @joshcooper I don't see any side effects with this. I'm going to merge it to keep things moving. If you see this as breaking something, please comment.

McdonaldSeanp:
Yeah, So the ticket spawned from a workflow where a fresh agent install on a fresh machine expects that if you don't pass the environment variable to the installer the entry won't exist

jpogran:
Validated with the following:

Brand new install:

```
PS C:\Users\vagrant> cat C:\ProgramData\PuppetLabs\puppet\etc\puppet.conf
[main]
server=puppet
autoflush=true
```

Upgrade from 1.4.1 to this

```
PS C:\Users\vagrant> puppet -V
4.4.1
PS C:\Users\vagrant> cat 'C:\ProgramData\PuppetLabs\puppet\etc\puppet.conf'
[main]
server=puppet
autoflush=true
environment=production
PS C:\Users\vagrant> msiexec /i C:\vagrant\puppet-agent-1.4.2.407.gb3c4030-x64.msi /qn
PS C:\Users\vagrant> puppet -V
4.5.0
PS C:\Users\vagrant> cat 'C:\ProgramData\PuppetLabs\puppet\etc\puppet.conf'
[main]
server=puppet
autoflush=true
environment=production
```

ferventcoder:
Doesn't remove it. We wouldn't want it to. @McdonaldSeanp something to keep in mind.

McdonaldSeanp:
Thanks.

PR #143, opened 2016-05-05
(RE-6903) Brand update!

McdonaldSeanp:
It's time to update the ui resources for the installer, and now that we have
the resources it's time to update before 1.5.0


mckern:
mmm, bitmaps. 👍

PR #145, opened 2017-04-10
Update the PATH order for puppet.bat

zymzxq:
Change the PATH order to ensure that `%PL_BASEDIR%\bin\puppet.bat` has the priority over `%PUPPET_DIR%\bin\puppet` so when we call `puppet` from the command-line, the bat file gets called.


joshcooper:
Thanks @zymzxq. Unfortunately, this repo is deprecated, as it is used to build 3.x version of puppet, which is no longer supported.

Could you open a "PA" ticket in tickets.puppet.com and submit a PR to the `puppet-agent` repo, see https://github.com/puppetlabs/puppet-agent/blob/master/resources/files/windows/environment.bat.

jcoconnor:
@zymzxq Can we close this PR - as @joshcooper indicated, a new ticket/PR needs to be raised against Puppet-Agent (PA)

